PR #2, opened 2021-05-25
Initial generation run

hsolbrig:


hsolbrig:
See edits to CONFIG.yaml -- what isn't clear in the documentation is that the organization and model name need to be "alliance-genome" (as above) and "agr_curation_schema".   I went ahead and ran  the make -f MakeConfig followed by make.  For some reason, the SQLDDL didn't generate - we need to talk about this...

PR #13, opened 2021-06-28
minimal changes required to get "make install all" working

gildossantos:
For `core.yaml`:
- Generic "Entity" created in core since it was referenced as parent of "OntologyTerm".

For `ontologyTerm.yaml`:
- Made `xrefs` slot a child of `cross_references` slot, imported from `crossReference.yaml`

For `geneInteraction.yaml`:
- Made "aggregation_database_enum" for the "aggregation_database" slot to get things working - not sure if there's a better way.
- Fixed typo in reference to `related_to` predicate.

For `orthology.yaml`:
- Removed redundant creation of `predicate` slot.

For `phenotypeAndDiseaseAnnotation.yaml`:
- Typos

### DIFF
--- a/model/schema/core.yaml
+++ b/model/schema/core.yaml
@@ -51,12 +51,19 @@ types:
 
 classes:
 
+  Entity:
+    abstract: true
+    description: >-
+      A generic database entity.
+    slots:
+      - id
+
   BiologicalEntity:
+    is_a: Entity
     description: >-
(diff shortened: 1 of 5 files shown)


sierra-moxon (approved):

sierra-moxon:
I made a small change to remove 'Entity' as a parent

PR #21, opened 2021-07-12
Docs delete

chris-grove:
@sierra-moxon    Please review
Note: Mac systems cannot distinguish files that are the same name except for case. So some artifacts in docs like:

~/docs/Allele.html
~/docs/allele.html

Cannot be considered different files on a Mac (pointed out by @oblodgett )
Olin helped me clean this up and make the commit.


oblodgett:
In the unix world Allele.html and allele.html are two different files. However in the mac world they are the same file. If someone on a unix / windows machine commits these, then what happens on the mac via git checkout / close / etc is the first file gets written to disk and then the second file overwrites the first file because "they are the same file" so it looks like the first file was modified however its now the second file. And what can be even more strange is the mac will show you two "different" files on the filesystem. The same physical file with "two pointers". Long story short... avoid this if possible.

sierra-moxon (approved):

sierra-moxon:
@oblodgett - indeed.  in fact, no one except github-actions user should be regenerating docs (we'll get there).

PR #28, opened 2021-09-01
Update phenotypeAndDiseaseAnnotation.yaml

chris-grove:
Consolidated PhenotypeAnnotationCurated and PhenotypeAnnotationInferred. Consolidated DiseaseAnnotationCurated and DiseaseAnnotationInferred. Added additional slots in the DiseaseAnnotation class definition (and slot definitions) to accommodate the attributes declared in the disease association file specification (DAF spec)

### DIFF
--- a/model/schema/phenotypeAndDiseaseAnnotation.yaml
+++ b/model/schema/phenotypeAndDiseaseAnnotation.yaml
@@ -36,7 +36,7 @@ prefixes:
 
 classes:
 
-  PhenotypeAnnotationCurated:
+  PhenotypeAnnotation:
     is_a: Association
     mixins:
       - AuditedObject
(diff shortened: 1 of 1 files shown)


model/schema/phenotypeAndDiseaseAnnotation.yaml
```diff
@@ -121,6 +116,18 @@ classes:
       - with
       - evidence_code           # added to avoid the need for a publication join
       - date_assigned           # dateAssigned in JSON schema
+      - inferred_gene
+      - inferred_gene_association
```
oblodgett, line 120: @chris-grove just wondering what the type is for this `inferred_gene_association` I would guess that if we have the association we might not need the inferred_gene as it would come with the association. Same with Allele?

chris-grove:
@oblodgett  I just tried to clarify in a new commit

PR #30, opened 2021-09-14
Disease subclass

sierra-moxon:
Break out disease annotation into geneDisease, alleleDisease, AGMDisease as an attempt to isolate kind-specific slots with their specific annotation type.  Also add back in constraints from ingest schema (ie: is_model_of only used with AGMDisease annotations).  

For discussion in DQM next week :)


chris-grove:
Thanks @sierra-moxon 
Here are my comments:

- 'sgd_strain_background' slot should be declared in GeneDiseaseAnnotation class, not the AlleleDiseaseAnnotation class; question to consider: Does SGD intend to use this slot for anything other than gene-disease annotations?
- 'with' slot should be declared in parent class DiseaseAnnotation if RGD intends to use it for strains (as is currently being submitted); if only SGD is using it for gene-disease associations then maybe it should instead be declared in the GeneDiseaseAnnotation class; in any case, it shouldn't be restricted to the AlleleDiseaseAnnotation class.
- 'genetic_modifier' and 'genetic_modifier_relation' needs to be declared in the parent class DiseaseAnnotation and not just in AlleleDiseaseAnnotation; Maybe we need 'gene_genetic_modifier', 'allele_genetic_modifier' and 'agm_genetic_modifier' slots, each with an appropriate range: Gene, Allele, or AffectedGenomicModel?
- In 'AGMDiseaseAnnotation' class definition, in slot_usage the subject should have range "AffectedGenomicModel" not "Allele" and the predicate range should be 'agm_disease_relation_enum' not 'allele_disease_relation_enum'

PR #43, opened 2021-11-11
added controlled vocabulary yaml

sbello:
Added file with the controlled vocabulary LinkML model
@chris-grove would you please review and let me know if it needs any corrections

### DIFF
--- a/model/schema/ontologyTerm.yaml
+++ b/model/schema/ontologyTerm.yaml
@@ -15,9 +15,7 @@ prefixes:
 slots:
   definition:
     description: >-
-      The definition of the ontology term.  This is a reference to an object that holds
-      the text description of the term and a collection of URLs that further define the
-      term.
+      The explanation of the meaning of a term.
     range: string
   display_synonym:
     description: >-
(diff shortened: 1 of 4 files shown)


chris-grove (approved):
Thanks @sbello !
I've cleaned up a few things here and there, ran 'make all' locally and everything seems to be in order.

chris-grove:
@sierra-moxon 
Could you just take a look at the header for this new 'controlledVocabulary.yaml' file and see that is looks OK?

model/schema/ontologyTerm.yaml
```diff
@@ -56,6 +54,12 @@ classes:
     defining_slots:
       - curie
       - namespace
+    slot_usage:
+      - definition:
```
sierra-moxon, line 58: - definition doesn't need the dash.

sierra-moxon:
looks great - just make sure to add the new yaml file to allianceModel.yaml -- as that is the "root" of our project that is used by all the generators to traverse our yaml files. 

https://github.com/alliance-genome/agr_curation_schema/blob/main/model/schema/allianceModel.yaml

chris-grove:
@sierra-moxon 
OK, thanks for catching those! Running 'make all' now to check and will merge after that.

PR #44, opened 2021-11-16
Updated slot descriptions, added new slots and classes for phenotype …

chris-grove:
…and disease annotations

- Changed interaction context 'data_provider' to 'interaction_data_provider'
- Added "PhenotypeTerm" subclass of OntologyTerm; removed "Phenotype" class
- Updated PhenotypeAnnotation class to align with approach for DiseaseAnnotation class
- Updated "with" slot description URL and slot_usage description in DiseaseAnnotation class
- Updated "produced_by" slot in AuditedObject to "created_by" to match "creation_date"
- Removed 'references' slot from top-level Association class (inheriting classes like DiseaseAnnotation have single-valued 'annotation_reference' slot now)

### DIFF
--- a/model/schema/core.yaml
+++ b/model/schema/core.yaml
@@ -132,7 +132,7 @@ classes:
       and the agent responsible for the change.
     slots:
       - table_key
-      - produced_by
+      - created_by
       - creation_date
       - modified_by
       - date_last_modified
@@ -162,12 +162,10 @@ classes:
     notes:
       Keeping this more generic than biolink:association by foregoing the
       qualifiers and negated slots.
-      object
     slots:
       - subject
       - predicate
       - object
-      - references
 
   EntitySynonym:
     is_a: Association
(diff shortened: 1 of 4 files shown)


chris-grove:
@sierra-moxon @markquintontulloch 
Thanks! I needed to comment out the phenotype relation enums, but now 'make all' runs fine. I'll merge now.

sbello:
Some questions on the phenotype section that we may need to bring to the working group:

Do we have this for phenotypes?
 predicate:
        description: The relationship between AGM and phenotype.
        range: agm_disease_relation_enum   # TO DO: generate this relation enumeration or don't specify predicate
        required: true
We don't currently on the Alliance pages and MGI doesn't.

Also the name of the enum should be changed to phenotype unless you were thinking we would reuse the disease relationship set?

How do the post composed annotations fit in this model?

For allele annotations:
    slot_usage:
      subject:
        description: >-
          The gene to which the phenotype ontology term is associated.
        range: Allele
In the description gene should be allele.

chris-grove:
Thanks @sbello !
I've cleaned those up here:
https://github.com/alliance-genome/agr_curation_schema/pull/47

PR #46, opened 2021-11-17
Initial commit of SequenceTargetingReagent schema

markquintontulloch:
Inherits cross_references from GenomicEntity slot.  Frozen JSON schema has singular modCrossReference field, but I don't see any reason to override the multivalued nature of the inherited slot in this schema.


chris-grove:
@markquintontulloch 
Thanks for starting this. I think I'd like to discuss this model with other curators. I'm not entirely convinced that sequence targeting reagents ought to be considered necessarily a "GenomicEntity" and thus also a "BiologicalEntity" and thereby require a species/taxon specification. To me it seems this class should more appropriately inherit from the "Reagent" mixin and be treated as such. From your current proposal I gather that these are the attributes of the SequenceTargetingReagent class:

curie (from BiologicalEntity)
taxon (from BiologicalEntity)
name (from GenomicEntity)
synonyms (from GenomicEntity)
cross_references (from GenomicEntity)
secondary_identifiers (from GenomicEntity)
so_term (from SequenceTargetingReagent)
target_genes (from SequenceTargetingReagent)

I agree that we need curie, name, synonyms, so_term and target_genes. If we don't automatically inherit the 'cross_references' and 'secondary_identifier' attributes from a parent class, I would want to check that these fields are actually used before instantiating/requiring them in the model. I think we could have 'taxon' but maybe it shouldn't be required. I think I need to discuss with @ybradford and other curators about how taxon is currently intended to be used. At WormBase, we intend to generate sequence targeting reagents moving forward (we don't currently) but these would represent clones frozen as plasmids in solution or in E.coli strains used to propagate the plasmid and are not inherently part of, for example, the C. elegans species.

chris-grove (changes requested):
Comments made above. I think this requires a bit more discussion before finalizing the model

markquintontulloch:
Sure - this first pass was all based on the current ingest schema ([sequenceTargetingReagent.json](https://github.com/alliance-genome/agr_schemas/blob/master/ingest/affectedGenomicModel/sequenceTargetingReagent.json)).  In that, taxon is required.

chris-grove:
Also, as I mentioned in daily scrum, we should probably add in a 'data_provider' or 'data_providers' attribute (both exist in Neo4J; I'm not exactly sure why).

markquintontulloch:
I've added data_provider (required).  I'll leave any further changes until after discussion with the SMEs.

sierra-moxon (approved):

PR #48, opened 2021-11-22
PaperHandle class, paper_handle slot, other edits

chris-grove:
- Added PaperHandle class and paper_handle and handle slots
- Declared 'reference' (single-valued) slot in core (distinct from 'references' slot)
- Changed 'annotation_reference' to just 'reference'
- Declared 'annotation_type' and 'genetic_sex' not required in DiseaseAnnotation class
- Fixed model syntax error

### DIFF
--- a/model/schema/core.yaml
+++ b/model/schema/core.yaml
@@ -492,8 +492,13 @@ slots:
 
   references:
     singular_name: reference
-    description: holds between an object and a reference
+    description: holds between an object and a list of references
     multivalued: true
+    range: Reference
+
+  reference:
+    description: holds between an object and a single reference
+    multivalued: false
     range: Reference
 
   original_reference:
--- a/model/schema/phenotypeAndDiseaseAnnotation.yaml
+++ b/model/schema/phenotypeAndDiseaseAnnotation.yaml
@@ -305,7 +305,7 @@ classes:
       - condition_gene_ontology           # geneOntologyId in JSON schema
       - condition_taxon                   # NCBITaxonId in JSON schema
       - condition_chemical                # ChemicalOntologyId in JSON schema
-      - paper_handle
+      - paper_handles
     slot_usage:
       condition_class:
         required: true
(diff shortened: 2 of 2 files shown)


chris-grove:
I'm getting a series of errors when trying to run 'make all' locally, but the errors seem unrelated to the content of my LinkML edits. Trying to sort out what's going on now.

chris-grove:
OK, 'make all' ran successfully locally once I published my branch

markquintontulloch (commented):
I’m not sure the PaperHandle class fulfils the needs of the ZFIN curation system.  Was the handle not used to label a set of experimental conditions rather than a paper?  There could be multiple sets of conditions within a single paper, and a handle linking to a paper gains little over using the paper ID directly.  I would have thought that the handle would want to link to either the ConditionRelation or the ExperimentalCondition class.

markquintontulloch:
@chris-grove  Ignore the above comment, or at least parts of it (never try checking a pull request on a phone!).  Wouldn't a simple _handle_ slot in the _ExperimentalCondition_ (or possibly _ConditionRelation_ class) fulfil the requirement here?  I don't think we would want to link to a paper for this (the same handle could be used for referencing conditions that are the same in multiple papers, or multiple handles could come from the same paper).  If my understanding is correct, then essentially this handle is just a convenience method to avoid having to remember or enter a curie/name.  Is this even something that we want to model or persist in the database, or is this something that only persists during a user session?

chris-grove:
@markquintontulloch 
Hi Mark, the issue is that ZFIN has these handles with distinct namespaces for each paper. So, the same experimental condition could have many different handles, each paper with its own distinct handle for a given experimental condition. This is how they do things now and they would like to continue using these handles and propagate their paper-specific handles to the Alliance. These would persist in the database (not a session specific handle).

markquintontulloch:
Okay, makes sense.  In that case, shouldn't the _paper_handle_ slot be a multivalued _paper_handles_ slot?

chris-grove:
@markquintontulloch 
Yes, good catch! Thanks! I've made the change and commit.

markquintontulloch (approved):

PR #52, opened 2021-12-09
Introduces higher level "Expression Experiment" that groups "ExpressionAnnotation" objects

gildossantos:
This change came out of Expression WG discussions Aug 2021.


sierra-moxon (approved):

model/schema/expression.yaml
```diff
@@ -12,14 +12,33 @@ imports:
...
+      - references
+      - biological_entity_assayed
+      - assay_used
+      - reagents_used
+      - specimen_genomic_model
+      - specimen_alleles
+      - condition_relations
+    slot_usage:
+      references:
+        multivalued: false    # Should I call this slot just "reference" then?
```
sierra-moxon, line 34: We need a name for reference when it is single-valued, which is different than 'Reference' (the name of the Class).  What about 'attribution'?

model/schema/expression.yaml
```diff
@@ -30,115 +49,126 @@ classes:
...
+          But, sometimes we know the figure but don't have the image.
           Optional, as existing annotations from some MODs have no related
           images or figures.
         required: false
-        multivalued: false
+        multivalued: true
       has_reference:
         multivalued: false
 
+  TemporalContext:
```
sierra-moxon, line 78: As a class, this would turn into a table of unique TemporalContexts - is this what we want?  Would a curator search for a unique combo of stage, age and temporal qualifiers?

model/schema/expression.yaml
```diff
@@ -30,115 +49,126 @@ classes:
...
+    description: >-
+      The developmental stage and/or age of the specimen in an annotation.
+    slots:
+      - developmental_stage
+      - age
+      - temporal_qualifiers
+
+  AnatomicalSite:
```
sierra-moxon, line 86: should anatomical site have a reference to an expression experiment?

model/schema/expression.yaml
gildossantos: I don't know if a curator would search like this, but at FB, these combinations do get re-used often for different expression annotations, so it was an attempt to normalize the data.

model/schema/expression.yaml
gildossantos: The "expressionAnnotation" has a "where_expressed" slot that points to "AnatomicalSite", and I think that is correct. I don't think that the higher level "expressionExperiment" object should have a reference to any anatomical sites or temporal terms.

gildossantos: e.g., an expressionExperiment could have many annotations, those annotations could span different stage/locations.

model/schema/expression.yaml
sierra-moxon: Does the temporal_qualifiers make this harder to reuse?  tracing down to temporal_qualifier_set, it seems like this should be at the ExpressionAnnotation level vs. the TemporalContext level?

gildossantos: Looking at this again, I think the temporal qualifier slot should be removed. FlyBase is the only MOD to have temporal qualifiers, but it turns out there's only one such term ("precursor"), and it's only ever used to qualify anatomical terms: e.g., "precursor" of "atrium".
So I guess the question is if it's useful to have a TemporalContext entity that combines only stage and age. Perhaps not.

model/schema/expression.yaml
```diff
@@ -12,14 +12,33 @@ imports:
...
+      Defined by the gene of interest, the specimen, the assay, the reagents 
+      (Antibody, Probe), and the reference. It groups ExpressionAnnotations.
+    notes: >-
+      Specimens are defined by their AGM and/or alleles (or gel lane at MGI).
     slots:
-      - anatomical_structure
-      - anatomical_substructure
-      - cellular_component
-      - anatomical_qualifiers
+      - curie
```
cmungall, line 24: I am glad to see CURIEs/prefixed ids used universally, but I like naming fields by their roles rather than types. Might it make the schema marginally more approachable if the slot was called "unique_identifier" or simply "id"?

(it would still be typed as a CURIE)

Very minor and if the ship has sailed on this, ignore

sierra-moxon (approved):
I think its is good to merge in now, we can make changes in subsequent PRs. :)

PR #53, opened 2021-12-13
SCRUM-879 vocabulary changes

markquintontulloch:
Add unique constraint to 'name' slot of 'Vocabulary' class.
Add 'abbreviation' slot to 'VocabularyTerm' class.
Move 'abbreviation' slot to core.yaml as now referenced by EcoTerm and VocabularyTerm.


oblodgett:
@markquintontulloch do you want to commit the `agr_curation_schema` directory? I think that is suppose to be for artifacts only?

sierra-moxon:
we should probably review our .gitignore.

markquintontulloch:
Will redo PR without all the extra files

PR #56, opened 2022-01-04
Fixing 'reference' slot 'Reference' class naming clash on Macs

chris-grove:
Macs have a problem with LinkML entities (e.g. classes or slots) have the same name except for case. We recently have added a 'reference' slot in addition to the existing 'Refefence' class which causes a naming conflict on Mac computers. I have changed the 'reference' slot back to 'annotation_reference' and added an 'entity_reference' slot to be used by the PaperHandle class and any other non-association/non-annotation entity going forward.

### DIFF
--- a/model/schema/core.yaml
+++ b/model/schema/core.yaml
@@ -496,8 +496,13 @@ slots:
     multivalued: true
     range: Reference
 
-  reference:
-    description: holds between an object and a single reference
+  annotation_reference:
+    description: holds between an annotation and a single reference
+    multivalued: false
+    range: Reference
+
+  entity_reference:
+    description: holds between an entity and a single reference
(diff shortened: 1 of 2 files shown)


markquintontulloch (commented):
It's not clear to me why we need separate annotation_reference and entity_reference slots.  Don't we just need a single non-multivalued reference slot that is called something other than 'reference'?

gildossantos (approved):

chris-grove:
@markquintontulloch 
Essentially, yes, but I don't know what that would/should be so I figured I could reinstate the 'annotation_reference' slot as we had it earlier. Maybe we should just have a single 'single_reference' slot or something. Thoughts?

PR #60, opened 2022-01-13
SCRUM-1037 More curie to unique_id changes

markquintontulloch:
Given the change from curie to unique_id on the DiseaseAnnotation class, I assume that this should also be done for the ConditionRelation and ExperimentalCondition classes which also a compound key (at the moment) rather than a curie.  This would avoid DQMs having to generate curies for these entities.


chris-grove (approved):

chris-grove:
@markquintontulloch 
Yep, sounds good!

PR #63, opened 2022-01-19
SCRUM-1068 Instantiate NCBITaxonTerm and update references

markquintontulloch:


model/schema/phenotypeAndDiseaseAnnotation.yaml
```diff
@@ -352,9 +352,7 @@ classes:
       condition_taxon:
         description: >-
           NCBITaxon ontology id used in subset of condition types like 'bacterial infection'.
-        range: OntologyTerm
-        values_from:
-          - NCBITaxon
+        range: NCBITaxonTerm
```
chris-grove, line 355: actually don't need this range constraint as it is redundant with the condition_taxon slot definition constraint (probably my own oversight when first writing the model)

chris-grove (changes requested):
In addition to the comments in-line:

In the antibody.yaml, there is an 'antigen_taxon' slot that still points to a range of 'uriorcurie'; this needs to update to 'NCBITaxonTerm'.

In the core.yaml, there is a 'from_species' slot (not used) that points to a 'Species' class (not used) as a range, and the 'Species' class has no definition (stub). I think we should delete the 'from_species' slot and the 'Species' class. Also in the core.yaml there is a slot definition for 'taxon_id' (not used) with description 'NCBI taxid' and range 'integer'. This 'taxon_id' slot should also be deleted.

markquintontulloch:
@chris-grove - agreed, I'll make the above changes.  

There are other 'range' redundancies that can be removed.  At the moment, the required status for the experimental condition slots is declared in the class - can these also be moved to the slots or is there potential for these to be used (and non-required) elsewhere in the model.  

Finally, I'm trying to work out what to do with condition_chemical.  This currently has a range of OntologyTerm, but the Molecule class is not an OntologyTerm.  We could create a 'Chemical' class that both CHEBITerm and Molecule inherit from, but this would stop CHEBITerm from being an OntologyTerm and I'm not sure if that makes sense (either from a logical or a database point of view).

markquintontulloch:
@chris-grove 

So if I understand correctly, the proposal is to create a ChemicalOntologyTerm class which is an OntologyTerm and would have the various fields that are currently preset on the Molecule class, such as iupac, inchi, etc..  Then, both Molecule and CHEBITerm classes would inherit from ChemicalOntologyTerm.  Is that right?

I think that makes sense.  We can load the iupac, inchi, etc. from the CHEBI ontology file.  The only minor issue I see is that the Molecule class entries would inherited some fields from the OntologyTerm class that are not applicable (although also not required), such as namespace, definition, definitionUrls.

Or have I got the wrong end of the stick?

markquintontulloch:
@chris-grove 

I've pushed some changes that should address all your comments.  I'll leave the conditionChemical issue for another PR as that is unrelated to SCRUM-1068.

chris-grove (approved):

PR #64, opened 2022-01-20
Added bulkload model

oblodgett:


markquintontulloch (approved):

model/schema/bulkload.yml
```diff
@@ -0,0 +1,198 @@
+id: https://github.com/alliance-genome/agr_curation_schema/src/schema/bulkload.yaml
+
+imports:
+  - core
+
+classes:
+
+  BulkLoadGroup:
+    mixins:
+      - AuditedObject
```
chris-grove, line 10: I realize that the AuditedObject class requires a 'created_by' and a 'modified_by' entry; will these always be available for BulkLoad, BulkLoadGroup, and BulkLoadFile objects? If yes, then OK, but if not then we either need to relax the requirements or something else.

model/schema/bulkload.yml
```diff
@@ -0,0 +1,198 @@
...
+    slots:
+      - backend_load_type
+
+slots:
+
+  loads:
+    multivalued: true
+    range: BulkLoad
+
+  status:
```
chris-grove, line 77: I think we're going to need a more specific label for this slot as 'status' is likely going to apply to many other classes. How about 'load_status'?

model/schema/bulkload.yml
```diff
@@ -0,0 +1,198 @@
...
+    range: bulk_load_status_enum
+
+  error_message:
+    description: Error message string if an error occurs
+    range: string
+
+  backend_bulk_load_type:
+    range: backend_bulk_load_type_enum
+
+  ontology_type:
```
chris-grove, line 88: 'load_ontology_type'? We may want an 'ontology_type' slot for the Ontology class at some point

model/schema/bulkload.yml
```diff
@@ -0,0 +1,198 @@
...
+    range: ontology_bulk_load_type_enum
+
+  group:
```
chris-grove, line 91: 'bulk_load_group'? Again, to avoid generic sounding slot names that might want to be instantiated later in a broader context.

model/schema/bulkload.yml
```diff
@@ -0,0 +1,198 @@
...
+
+  cron_schedule:
+    description: A string describing the cron syntax for the schedule
+    range: string
+
+  scheduling_error_message:
+    description: Any errors in syntax on the cron_schedule
+    range: string
+
+  data_type:
```
chris-grove, line 134: 'fms_data_type'?

model/schema/bulkload.yml
```diff
@@ -0,0 +1,198 @@
...
+    description: The dataType paramater in the FMS
+    range: string
+  data_sub_type:
```
chris-grove, line 137: 'fms_data_sub_type'?

model/schema/bulkload.yml
```diff
@@ -0,0 +1,198 @@
...
+    description: The dataSubType paramater in the FMS
+    range: string
+  url:
```
chris-grove, line 140: 'bulk_load_url'?

markquintontulloch: Does this need to be distinguished? A URL will always be a URL so this can be reused elsewhere.

chris-grove: Well, I'm fine with declaring a generic 'url' slot, but then it should probably be declared in the core.yaml and the description in the generic slot definition should not be so specific as 'The URL that this bulk load will download for ingest'. Such a specific comment would better fit in slot_usage so as not to force it to apply to all situations in which someone wants to use a 'url' slot. We might also want the range constrained more specifically than just 'string' but rather 'uriorcurie' or something.

chris-grove (changes requested):
Comments in-line

model/schema/bulkload.yml
oblodgett: They will be set via whoever is logged into the UI.

model/schema/bulkload.yml
oblodgett: Ok changed to bulkload_status

model/schema/bulkload.yml
oblodgett: We could go down the class road. However we are more specifying a type which we do a lookup on the backend to get the class.

model/schema/bulkload.yml
oblodgett: Changed to bulkload_group

model/schema/bulkload.yml
oblodgett: Changed to fms_data_type

model/schema/bulkload.yml
oblodgett: Changed to fms_data_sub_type

model/schema/bulkload.yml
oblodgett: Changed to bulkload_url

chris-grove (approved):

PR #66, opened 2022-01-27
SCRUM-1079: adding slots to Person

abecerra:


markquintontulloch (changes requested):
In your slot definitions, 'notes:' should be 'description:'.

We already have a mod_id slot defined in the phenotypeAndDiseaseAnnotation.yaml.  You could move the slot definition from that file into core.yaml and remove the definition from this file.  The mod_id domain would be left unspecified as it would be used by multiple classes.  The range is currently set as string, but maybe this should be uriorcurie.

Just a stylistic point:  the '>-' is only required if the following text is not on the same line.

markquintontulloch (changes requested):
Also, unique_id is already defined in core.yaml, so you can remove the slot definition here.  

first_name and last_name are currently defined in informationContentEntity.yaml.  You should move these to core but you will need to remove the domain, and I believe the range should be string.

azurebrd (commented):
Hi @abecerra and @chris-grove  .  I'm not sure how far along the Person modeling is, but ideally they would also have a string middlename field, and the email field would allow a list.  WormBase at least doesn't track what the primary email is, we try to put the most current email in the first slot in the curation DB, but I don't think that makes it into acedb in a sorted way.  We haven't discussed with other mods how they handle email primary vs other.

abecerra:
> In your slot definitions, 'notes:' should be 'description:'.
> 
> We already have a mod_id slot defined in the phenotypeAndDiseaseAnnotation.yaml. You could move the slot definition from that file into core.yaml and remove the definition from this file. The mod_id domain would be left unspecified as it would be used by multiple classes. The range is currently set as string, but maybe this should be uriorcurie.
> 
> Just a stylistic point: the '>-' is only required if the following text is not on the same line.

Thank you Mark! 1) Changed notes per description, 2) Removed '>-' instances, 3) Moved slot defs to core

abecerra:
> Also, unique_id is already defined in core.yaml, so you can remove the slot definition here.
> 
> first_name and last_name are currently defined in informationContentEntity.yaml. You should move these to core but you will need to remove the domain, and I believe the range should be string.

Moved with domain and range changed

abecerra:
> Hi @abecerra and @chris-grove . I'm not sure how far along the Person modeling is, but ideally they would also have a string middlename field, and the email field would allow a list. WormBase at least doesn't track what the primary email is, we try to put the most current email in the first slot in the curation DB, but I don't think that makes it into acedb in a sorted way. We haven't discussed with other mods how they handle email primary vs other.

Thank you Adam, I added slots middle_names and other_emails

markquintontulloch (approved):
The changes I requested look good.  I think this is fine as a first pass.  There are possibly a couple of outstanding questions that may require input from others:
1.  Given WB (and possibly others) cannot identify which stored email is the primary one, perhaps instead of adding the `other_emails` slot, `email ` should become `emails` with `multivalued: true`
2. I personally don't like the plural slot name `middle_names` when `multivalued: false`.  On the Java side I would expect any plural variable name to correspond to a list.  A think a list for middle names would cause headaches due to the requirement to keep it ordered, so I would opt for `middle_name` with `multivalued: false`, although obviously multiple names in a single string would be a permitted value

abecerra:
> The changes I requested look good. I think this is fine as a first pass. There are possibly a couple of outstanding questions that may require input from others:
> 
> 1. Given WB (and possibly others) cannot identify which stored email is the primary one, perhaps instead of adding the `other_emails` slot, `email ` should become `emails` with `multivalued: true`

Yes, probably

> 2. I personally don't like the plural slot name `middle_names` when `multivalued: false`.  On the Java side I would expect any plural variable name to correspond to a list.  A think a list for middle names would cause headaches due to the requirement to keep it ordered, so I would opt for `middle_name` with `multivalued: false`, although obviously multiple names in a single string would be a permitted value

I was tempted to remove the plural, but I thought that maybe it was there for a reason

PR #67, opened 2022-01-28
SCRUM-1079: fixes for middle_name

abecerra:


oblodgett (approved):

azurebrd:
Hi @abecerra @oblodgett I'm not sure what drove this change.  If it was from what I said about not wanting middle_names multivalued, then thanks.  But the label middle_names was technically accurate, since it's a string field for a messy list of middle names.  That is, at the Mods, no one tokenizes the middle names into individual middle names to put in separate entries, and its because it can be hard to know whether two words are one middle name or two middle names (in the human name sense).  Either way, if this changes it to being singlevalued, that's great, thanks.  And it's fine with me if the label looks singular, contains a single string, but contains multiple human middle names.

@cerivs was asking me if this would break things with the reference JSON, and I don't know.  Does this affect the validator when DQMs upload files, or is this modeling for downstream when data gets parsed for the persistent databases ?

markquintontulloch:
@azurebrd - the change to a singular name `middle_name` was motivated by the fact that pluralised names imply that a list is expected on the Java code side.  This change was to make it clear that a single string (albeit possibly one consisting of multiple names) is expected.

In regards to the question from @cerivs - this will have no immediate bearing on the reference JSON.  That is based on the current frozen (1.0.1.4) schema.  The LinkML-derived schema is distinct from the old system - it being the schema that will be used for ingest into the new curation database.  The literature system API should generate JSON that matches this new schema to ensure smooth interoperability between the curation system and the literature system (that is the point of LinkML after all), so presumably there will be a change to the reference JSON schema at some future point to avoid having to convert between the two.

azurebrd:
Awesome, thanks @markquintontulloch :)   That all makes sense, and is great.

PR #68, opened 2022-01-29
change 'reference' slot to 'single_reference'

chris-grove:


azurebrd (commented):
I don't know if this is preliminary work, but some of these Date fields might be String fields, and not proper date fields.  In the Literature database they're being handled as Strings http://dev.alliancegenome.org:4003/redoc#operation/create_reference__post  I haven't parsed the data in months, so I don't recall what's in them.  Some of them might be able to be dates, but some will for sure not.  @chris-grove @sierra-moxon

PR #69, opened 2022-01-29
Grove Person changes 1-29-2022

chris-grove:
- Added LoggedInPerson class with new 'okta_id' and 'okta_email' slots
- Added domain of Person for first, middle, and last names
- Made orcid single valued
- Made 'emails' slot multivalued and changed description
- Add 'old_emails' slot definition and added slot to Person class


chris-grove:
Trying to add this commit to the same branch:
https://github.com/alliance-genome/agr_curation_schema/commit/d565f1c12b67c7666b2f6d1fa26c6f5f4b009cd6
This commit adds the 'old_emails' slot

chris-grove:
Trying to add this commit to the same branch:
https://github.com/alliance-genome/agr_curation_schema/commit/d565f1c12b67c7666b2f6d1fa26c6f5f4b009cd6
This commit adds the 'old_emails' slot

markquintontulloch (changes requested):
The AuthorReference class uses the various name slots, so we set the domain to Person.  This is why we put the slots in core and not with Person.  However, I’m not sure of the purpose of the AuthorReference class and it seems to me that it could either reference Person or be replaced by it.

azurebrd (commented):
Looks great, thanks @chris-grove

azurebrd:
Hi @markquintontulloch .  (My comment that it looked great was that it works well for how I'm envisioning a Person class, not that it's good for how LinkML handles other classes).  I'm not sure how AuthorReference is used in the LinkML context either.  I don't know if what I'm going to say here is already known and obvious to you, but in case it helps clarify:  Papers are Reference objects.  They have Authors associated with them (I don't know if they're the AuthorReference mentioned), these authors sometimes have ORCID from PubMed, also Last Name, and either or both First Name / First Initial, sometimes they'll have institution affiliation (They do not have emails from pubmed, but they might be extractable from the PDF, but it might be ambiguous which author an email corresponds to).  Separately we have Persons, which can be connected to the Authors, or to other Persons, or to curation entries, and probably other stuff.  These Persons have First name, Middle names, Last Name, ORCID, emails, address, institution, and a bunch of other stuff.  It helps to think of an Author as being part of a Paper/Reference, and a Person as being a clean + curated object that maps to multiple Authors.

chris-grove:
@markquintontulloch 
Ah OK, thanks! I've removed the Person domain constraints from first_name, middle_name, and last_name

abecerra (approved):

markquintontulloch (approved):

chris-grove:
@azurebrd @abecerra @markquintontulloch 
OK, thanks for the feedback! @oblodgett and I discussed and we figured we'll wait until the current A-Team sprint is over before merging this branch, FYI.

PR #74, opened 2022-02-09
SCRUM-1185: Add "Note" object.

gildossantos:
NB - I'm leaving EntityStatement in place now so that I can coordinate the replacement of it with the expression working group, but I wanted to get the new Note in place ASAP.

### DIFF
--- a/model/schema/core.yaml
+++ b/model/schema/core.yaml
@@ -326,6 +326,15 @@ slots:
       Default value is true.
     range: boolean
 
+  related_notes:
+    description: Holds between an object and a list of related Note objects.
+    notes: >-
+      The original name suggested for this slot was simply notes, but I think
+      that label is reserved in LinkML for internal descriptions of objects,
+      hence the related_notes label used here.
+    range: Note
+    multivalued: true
+
   statement_subject:
     description: >-
       The entity being described by the note.
--- a/model/schema/core.yaml
+++ b/model/schema/core.yaml
@@ -144,6 +144,22 @@ classes:
       - generated_by
       - manufactured_by
 
+  Note:
+    description: >-
+      Free-text describing some attribute of an entity.
+    slots:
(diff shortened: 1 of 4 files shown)


model/schema/core.yaml
```diff
@@ -293,6 +309,32 @@ slots:
...
 
+  free_text:
+    description: >-
+      A free text string that describes some aspect of an entity.
+    range: string
+
+  note_type:
+    description: >-
+      The type of note: e.g., cytology, comment, summary.
+    range: OntologyTerm
```
chris-grove, line 320: The range of 'note_type' should be 'VocabularyTerm' from the controlledVocabulary.yaml, not 'OntologyTerm'; this also means that we'll need to add to the 'imports' list 'controlledVocabulary'

model/schema/core.yaml
```diff
@@ -144,6 +144,22 @@ classes:
       - generated_by
       - manufactured_by
 
+  Note:
+    description: >-
+      Free-text describing some attribute of an entity.
```
chris-grove, line 149: I would prefer a description like: "Note object for capturing free-text describing some attribute of an entity, coupled with a 'note type', private boolean, and an optional list of references." Basically, I want to make sure that we don't just define the Note object as only free-text.

chris-grove (changes requested):
There are two comments to address plus the addition of 'controlledVocabulary' to the list of core.yaml imports.

model/schema/core.yaml
gildossantos: Yes, that makes sense. Done.

model/schema/core.yaml
gildossantos: Sorry. Fixed now.

chris-grove (approved):
Looks good, thanks!

PR #77, opened 2022-02-10
Add WB disease annotation test file and Person class update

markquintontulloch:
Correct branch this time!

Identifier for Person class set as unique_id to prevent inlining in JSON schema.  Test files updated accordingly.
WB disease annotation test file added.


sierra-moxon (approved):

sierra-moxon (approved):

azurebrd:
Ah, thanks @markquintontulloch .  @cerivs and I were wondering what the unique_id was for, seems like it's for the Alliance Person ID, and if so, that's great.  I think Ceri is going to talk to Sierra about how ZFIN stores their curator evidence for Disease Annotations at ZFIN, which is a minor heads up, I know Sierra signed off on only getting the 'ZFIN' string.
I'm unclear about why reference became single_reference .  Is that so there can be single_reference objects that just refer to PMIDs, as opposed to a future integration of Alliance Reference IDs once Literature switches over to whatever ID format everyone agrees to ?  Blue has on the backlog to work on the Reference LinkML modeling, and I don't know whether that should be on the single_reference class, or the reference class. 
Thanks.

markquintontulloch:
@azurebrd - if I understood and remember the A-Team discussions correctly, for now unique_id will simply be whatever each MOD uses to uniquely identify a person.  This could be a MOD ID, email address, or some sort of generic string if a person cannot be identified.

The change to single_reference was simply because the slot name clashed with the Reference class - this causes build failures on Macs.  It still points to the Reference class, but is single-valued as opposed to the multivalued references slot.  The value in the JSON schema is just a string (as the slot is not 'inlined', the schema just requires a string rather than the full Reference object) but the value of this string should be the identifier of the class the slot points to, in this case the Reference curie.

PR #82, opened 2022-02-14
Add "related_notes" slot to "Antibody" class and delete the "AntibodyStatement" class.

gildossantos:
These suggested changes were necessitated by the implementation of the new "Note" class, which is intended to replace the "EntityStatement" class.

Even though they are no longer referenced by anything, I left the `antibody_note_type_set` enum there for future reference. I suspect these will need to become ControlledVocabularyTerm objects so that the proper Note types can be created.

### DIFF
--- a/model/schema/antibody.yaml
+++ b/model/schema/antibody.yaml
@@ -45,6 +45,7 @@ classes:
       - secondary_identifiers
       - references
       - original_reference
+      - related_notes
       # - disease    # Need a way to give DOID term and reference. Or a generic way to specify CV term and publication attribution.
     slot_usage:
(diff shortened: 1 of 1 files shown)


edwong57 (commented):
ok

draciti:
@edwong57 @gildossantos, does this have repercussions on the Expression model notes and statements?

gildossantos:
Yes @draciti, changes will be required, but I think you'll still have all the note types you want. They'll just modeled slightly differently on the backend.

draciti:
👍🏻👍🏻
________________________________

PR #85, opened 2022-02-15
Variant updates

markquintontulloch:
Initial rewrite of Variant class and related classes according to variants WG specification: [variants spec](https://docs.google.com/spreadsheets/d/1pdB7tuM8sNSu-vF1GaPmDAg0lxHGffr7AWtddUBznHI/edit#gid=266437103)

Further discussion required with variants WG on how to integrate VEP results, currently modelled in variantConsequence.yaml


sierra-moxon:
@markquintontulloch  - what if we had a typing slot we could use to disambiguate between "curated" and "source" instead of dual slots for each?  I can definitely see the benefit of being explicit (e.g. the interface might need to pull the source in right next to the curation and in my suggestion, the code would have to aggregate two "rows" rather than one). I worry a bit abstractly about duplication of 'columns'.   

Also, this is a fairly new addition to linkml, but I think we could make use of: 1) anyOf conventions to further validate/restrict the attributes available for each variant type, and 2) instead/in addition regex's rules to constrain.  Let me pull together that doc - edits to this comment forthcoming.

markquintontulloch:
@sierra-moxon - Thanks for the comments.  Yes, the idea of a typing slot had occurred to me.  However, I was unsure about how to deal with those VariantLocation slots that aren't duplicated for curated/source (i.e. `evidence_code` and `hgvs`).

The use of anyOf and regex rules are things I was unaware of in LinkML - they sound promising :)

Also, from the failing build I see that there is a clash between the `evidence_code` slots in my updated variation.yaml and orthology.yaml - the former has a range of an ECOTerm object, the latter an enum that is currently undefined.   Can we change the latter to also point to an ECOTerm object?

PR #88, opened 2022-02-22
Minor disease annotation updates

markquintontulloch:
data_provider and secondary_data_provider slots changed to multivalued=false
disease_qualifiers slot changed to multivalued=true
private slot renamed to internal

@gildossantos @edwong57 - the first of these changes will cause your test files to fail


gildossantos:
@markquintontulloch - I could see the `private` to `internal` change in the tests, but not in the `core.yaml` - so I just added them. Hopefully that is what was intended.

markquintontulloch:
@gildossantos - yes, thanks for catching that.

PR #89, opened 2022-02-22
Add rules examples

sierra-moxon:
rule example is [here](https://github.com/alliance-genome/agr_curation_schema/pull/89/files#diff-3ad08fceb1ae6b213a6928861837d185e21feb175747cf56cd1fe129116340ba)

### DIFF
--- a/model/schema/core.yaml
+++ b/model/schema/core.yaml
@@ -137,14 +137,6 @@ classes:
       - modified_by
       - date_last_modified
 
-  Reagent:
-    is_a: BiologicalEntity
-    description: >-
-      A material entity used in experiments.
-    slots:
-      - generated_by
-      - manufactured_by
-
   Note:
     description: >-
       Note object for capturing free-text describing some attribute of an entity,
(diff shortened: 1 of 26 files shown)


model/schema/expression.yaml
```diff
@@ -23,16 +23,24 @@ classes:
...
+      - postconditions:
+          any_of:
+            - slot_conditions:
+                biological_entity_assayed:
+                  required: true
+            - slot_conditions:
+                reagents_used:
+                  required: true
+          description: At least one of biological_entity_assayed or reagents_used should be populated.
+
```
gildossantos, line 43: Further down in the file, `reagents_used` is not listed as `required` but `biological_entity_assayed` is listed as `required`. Does that need to change?

model/schema/core.yaml
```diff
@@ -138,7 +138,7 @@ classes:
       - date_last_modified
 
   Reagent:
```
gildossantos, line 140: Does it make sense to move the `Reagent` stuff to the new `reagent.yaml` so all of that stuff is in one place?

gildossantos (approved):

model/schema/expression.yaml
sierra-moxon: yep! Good catch.

PR #92, opened 2022-03-01
Rename bulkload.yml to bulkload.yaml

chris-grove:
...to be consistent with the other yaml files in the directory and to match the id:

https://github.com/alliance-genome/agr_curation_schema/src/schema/bulkload.yaml


oblodgett (approved):

chris-grove:
@oblodgett  Now the checks are complaining of the range 'long':

sierra-moxon:
yes int

PR #97, opened 2022-03-07
update reference to conform to literature database

dougli1sqrd:
New PR into the correct repo! For https://agr-jira.atlassian.net/browse/SCRUM-1142

### DIFF
--- a/model/schema/core.yaml
+++ b/model/schema/core.yaml
@@ -439,7 +439,7 @@ slots:
     range: integer
 
   creation_date:
-    aliases: [ 'publication date' ]    # Can we revisit this? I don't think creation_date and 'publication date' are the same thing (CG)
+    aliases: [ 'date_created' ]
     description: >-
       The date on which an entity was created. This can be applied to nodes or edges.
     range: date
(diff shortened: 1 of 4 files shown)


azurebrd (commented):
date_last_modified_in_PubMed will eventually become date_last_modified_in_pubmed when we can change the database and API / data.  Right now our database calls it date_last_modified , which is not right.

dougli1sqrd:
@azurebrd are you saying there should be a slot `date_last_modified_in_pubmed` still?

gildossantos (approved):

azurebrd:
@dougli1sqrd There will eventually need to be a change from date_last_modified to date_last_modified_in_pubmed once the database and API change to match.  So, more of a heads up that it will change, but it hasn't yet.

gildossantos:
@azurebrd - I added a few notes in the LinkML that explains how currently the `date_last_modified` should be renamed to `date_last_modified_in_pubmed`.

azurebrd:
@gildossantos I see it, thanks Gil !

PR #98, opened 2022-03-08
SCRUM-1142 tweaks

gildossantos:
Some suggested modifications to the work on updating Reference LinkML

### DIFF
--- a/model/schema/informationContentEntity.yaml
+++ b/model/schema/informationContentEntity.yaml
@@ -81,22 +81,22 @@ slots:
     range: InformationContentEntity
 
   title:
-    multivalued: false
     description: >-
       A human readable title for a reference.
+    multivalued: false
     domain: InformationContentEntity
     range: string
 
-  volumes:
-    singular_name: volume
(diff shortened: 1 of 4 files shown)


model/schema/informationContentEntity.yaml
```diff
@@ -81,22 +81,22 @@ slots:
...
     description: >-
       A human readable title for a reference.
+    multivalued: false
     domain: InformationContentEntity
     range: string
 
-  volumes:
-    singular_name: volume
-    multivalued: false
+  volume:
```
gildossantos, line 90: column name in ABC is just "volume"

model/schema/reference.yaml
```diff
@@ -21,6 +21,7 @@ classes:
       - abstract
       - category
       - citation
+      - curie
```
gildossantos, line 24: Added because there is a `references.curie` column in the ABC db. Not sure if this was intentionally left out.

dougli1sqrd:
Thanks Gil!

PR #100, opened 2022-03-14
SCRUM-1329 model paper handle for experimental conditions

cmpich:
A PR for discussions.

### DIFF
--- a/model/schema/phenotypeAndDiseaseAnnotation.yaml
+++ b/model/schema/phenotypeAndDiseaseAnnotation.yaml
@@ -380,6 +380,8 @@ classes:
       conditions relevant to the annotation.
     slots:
       - unique_id
+      - handle
+      - single_reference
       - condition_relation_type
       - conditions
     slot_usage:
@@ -388,26 +390,14 @@ classes:
           Unique identifer for the condition relation.  Will be generated at
           AGR.
         required: false
+      handle:
+        required: false
+        multivalued: false
+      single_reference:
+        required: false
       condition_relation_type:
         required: true
         multivalued: false
-
-  PaperHandle:
-    description: >-
-      A pairing of a reference and a free text string that allows an object to have
-      a reference-specific alias (or handle). For example, used for experimental
-      conditions from ZFIN to label (in a reference-specific manner) individual
-      experimental conditions when curating a particular reference.
-    slots:
-      - single_reference
-      - handle
-    slot_usage:
-      single_reference:
-        required: true
-      handle:
-        required: true
-        multivalued: false
-
 
 ## ------------
 ## SLOTS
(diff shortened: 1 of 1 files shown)


markquintontulloch (commented):
There’s an inconsistency between ‘handle’ in the class definition and ‘paper_handle’ in the slots section.  I presume ‘handle’ is the correct version as it is no longer linked to a reference in the model.  My only concern is that there’s nothing in the model that would prevent curators from specifying  duplicate handles. Otherwise, looks good to me.

chris-grove:
@cmpich @sierra-moxon 
I agree with @markquintontulloch 
The 'paper_handle' slot definition seems like it needs to be changed to 'handle' (build currently fails otherwise).

With the proposal to just attach a handle to the ConditionRelation object, now it means that we could have the same set of conditions in the same paper with different paper handles because they are used with different condition relations. So, we could have two conditions 'high salt' and 'high temperature' that are used with the 'has_condition' relation in some paper, with a paper handle 'high salt, temperature (has_condition)'; If another experiment used the same two conditions but with a different relation, e.g. "induced_by", it would then need a new and different paper handle like 'high salt, temperature (induced_by)' or something. Is this what we want? 

I had thought ConditionRelation objects would get reused across different papers, but it sounds like that isn't necessary, in which case it is OK to reference a single reference/publication. I see that neither 'handle' nor 'single_reference' are required individually, but it seems that if one is provided, the other should be provided as well. Is there a way we can specify that?

sierra-moxon:
Hi @markquintontulloch, @chris-grove - I think the handle does have to take condition_relation_type into account, yes, if it is intended to function the way it does at ZFIN (as a shorthand name for a set of conditions).  But, I was under the impression that the handle would only be used for ZFIN?  Is that still true?  If so, a ZFIN curator does not currently make condition_relation_types other than 'has_condition.'  (the highsalt hightemp use case does not currently exist at ZFIN).   

I think we could have a conversation about what makes a ConditionRelation unique and whether or not the harmonized model leads to two different ways of curating?  If I understand correctly, for ZFIN, a DiseaseAnnotation will only every have 1 ConditionRelation, whereas other groups may have many ConditionRelations per DiseaseAnnotation.  This could lead to questions about what the "handle" means if used for other groups.  Do we have to address this use case now?  

I'm curious why single_reference isn't required?  I think it probably should be.  It might not be very frequent that two papers use the exact same set of experimental conditions?

cmpich:
Those are valid comments, Chris, however, I would go for the simplest solution at this moment which serves the needs of ZFIN. As we develop the UI for the DA and the experimental conditions we will see if other MODs would be amenable to re-use the handle concept as well. We then can make the necessary modifications according to the new requirements. They probably will only be stated with a functioning UI to start with.
Sierra, thanks for fixing the paper_handle -> handle issue.

chris-grove:
@sierra-moxon @cmpich 
OK, yes, the most important thing now is handling the ZFIN use case, so if the model does that as proposed then I think we should move forward with it. I'm OK with making single_reference required, as long as we don't end up in a situation where we need to update the same reference information in multiple places. In other words, the DiseaseAnnotation class already has a single_reference slot, and so it should go without saying that a ConditionRelation object connected to a DiseaseAnnotation object has that same DiseaseAnnotation single_reference. If we can ensure that (in the software logic) the single_reference in the DiseaseAnnotation and the single_reference in the associated ConditionRelation objects are never decoupled such that the reference information has to be managed in multiple locations, then I'm OK with it.

I will go ahead and merge as is. Thanks!

PR #105, opened 2022-03-28
Variant and associated class updates

markquintontulloch:
The aims of these changes are as follows:

1.  Enable multiple assemblies without excessive duplication of fields:  Assembly-independent fields are kept in the variant class.  A new assembly would result in a new row in the variantGenomeLocation table, linked to the same row in the variant table.  In cases where the transcript model simply lifts from one assembly to another, there would be no need for any additional rows in the variantTranscriptLocation table, but if the trancripts change between assemblies then a new row would be added to point to the new transcript (same goes for variantPolypeptideLocations and polypeptides).  
2. Separate source annotations from curated ones: The creation of the SourceVariantLocation class means that source annotations can be tied to a particular reference, and prevents unnecessary duplication of fields when creating an entry for a new assembly.
3.  Link VariantConsequence entities to VariantLocation entities, not Variant entities - if a variant can have multiple assemblies, we cannot link a consequence to the variant itself as this is assembly-specific.


sierra-moxon (approved):
looking great!  I like how this requires a genome location for a variant.  What do you think about making it even more explicit by breaking out variant_locations into its three components (genomic, transcript, and protein) and making genomic required?  Then I think we more or less have a definition of variant in the variant class that is consistent with our definitions?   
I hope I was able to fix the build issue - let's see if this builds now.

PR #109, opened 2022-03-31
add aliases

gildossantos:


azurebrd (approved):
That's great, thanks @gildossantos !

PR #112, opened 2022-04-04
SCRUM-1246 - LinkML for MOD association table

gildossantos:


azurebrd (commented):
Also recusing myself.  But Blue team pointed out corpus in modCorpusAssociation wasn't implemented as an enum, so I've changed the description and remove the enum.

gildossantos:
Thanks for fixing that @azurebrd!

model/schema/modCorpusAssociation.yaml
```diff
@@ -0,0 +1,87 @@
...
+    notes: >-
+      There should be only one corpus call for any given reference-mod pair.
+    slots:
+      - corpus
+      - date_created
+      - date_updated
+      - mod_corpus_association_id
+      - mod_corpus_sort_source
+      - mod_id
+      - reference_id
```
markquintontulloch, line 35: Is there a reason for the slot to just point to an ID and not the mod/reference object itself?

model/schema/modCorpusAssociation.yaml
```diff
@@ -0,0 +1,87 @@
...
+      For a given reference and Mod, whether it is inside corpus,
+      outside corpus, or needs review, as well as where this
+      sorting came from.
+    notes: >-
+      There should be only one corpus call for any given reference-mod pair.
+    slots:
+      - corpus
+      - date_created
+      - date_updated
+      - mod_corpus_association_id
```
markquintontulloch, line 32: Is this required in the model?  There is a curie slot inherited from the InformationContentEntity class.

model/schema/mod.yaml
```diff
@@ -0,0 +1,70 @@
...
+    is_a: Organization
+    description: >-
+      A Model Organism Database that is a member of the Alliance of Genome Resources.
+    notes: >-
+      We have not modeled synonyms, e.g. Mouse Genome Informatics
+      Literature has 8 values: FB, MGI, RGD, SGD, WB, XB, ZFIN, GO.
+    slots:
+      - abbreviation
+      - date_created
+      - date_updated
```
markquintontulloch, line 30: There are already date_last_modified and creation_date slots in the core.yaml.  I see from the reference.yaml that these are intentional aliases - is this just because it would be too hard at this point to rename these fields in the literature DB?

markquintontulloch (commented):
All of my comments essentially relate to the same issue.  I'm slightly confused by the use of IDs in the model.  I guess this is because you're working backwards from an existing DB and so the approach contrasts with how we've done things in the A-Team.  I don't feel particulary qualified to either approve or request changes to this PR - perhaps @sierra-moxon or @oblodgett can offer some more insightful comments/suggestions?

model/schema/modCorpusAssociation.yaml
gildossantos: We should probably delete the `is_a: InformationContentEntity` then to reflect the existing `mod_corpus_association` table.

model/schema/mod.yaml
azurebrd: @markquintontulloch Yeah, we've spent the first half of the release working to get things to the point where we can add a new API endpoint, and I think we're there just the day before review.  We only have the rest of the release to work on everything else that needs to happen for Literature, so we have renaming the tables to be singular instead of plural in the backlog.  Chris G requested we make the PubMed-based Biblio date fields proper date fields instead of strings, and I'm pushing for that to happen, as I've already parsed the data, but we can't even prioritize doing a fix for that (it's in the backlog).  Doing a renaming of the internal date_created / date_updated created by continuum only affects the LinkML and it probably isn't even in anyone's radar.

chris-grove:
@gildossantos 
I'm wondering if we can just add the new "Mod" class to the existing agent.yaml and the "ModCorpusAssociation" class to the existing reference.yaml, so as to avoid proliferation of .yaml files. Also, I think it would be fine to capitalize the MOD acronym in the class names as "MOD" and "MODCorpusAssociation" similar to how we have ontology term classes like "GOTerm" or "DOTerm", etc.

As for the slot "mod_id", I had originally added a "mod_id" for classes like DiseaseAnnotation to store the MOD's identifier for that data class object, like an individual disease annotation from WormBase like "WBDOannot00000001". The feedback was that "mod_id" sounded like the identifier for the MOD itself and so I changed the slot to "mod_entity_id". For the "mod_id" suggested here, I'm not entirely sure of the purpose. I suppose some of these slots are being created to match what is in the Literature system code?

PR #120, opened 2022-04-20
Adding test data for AGM, SqTR, Allele

sierra-moxon:
- also updated the required fields of Allele and AGM and SqTR (made name required).  I would've put this in GenomicEntity, but since Variant doesn't have a name, this seemed like a better choice.
- added references
- merged in PR 119 so that I could have the modeling components needed to verify the tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -182,6 +182,9 @@ SCHEMA_TEST_EXAMPLES := \
 	gene_test \
 	wb_disease_test \
 	variant_test \
+	agm_test \
+	sqtr_test \
+
 
 SCHEMA_TEST_EXAMPLES_INVALID := \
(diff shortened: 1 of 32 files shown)


sierra-moxon:
I am going to merge this in, as no one is submitting this data yet and this sets the stage for DQMs to look at some test data in prep to submit. 

Putting a note on the DQM agenda to talk about the Synonym class (no properties until this PR which adds a redundant synonym slot just to allow synonyms to be submitted -- otherwise with "additional properties" not allowed and no properties at all, we wouldn't be able to use this class.  I imagine this is an object because we want to talk about additional properties of a synonym -- maybe exact, narrow, broad, or something else).

PR #126, opened 2022-04-25
Make AuditedObject base class

chris-grove:
This PR is to make the AuditedObject class the base class for all other classes in the Alliance LinkML model.

Note that the "internal" boolean slot has been added to the AuditedObject base class so that all objects can be indicated as either internal = true (private) or internal = false (public) and has been indicated to be required, meaning that all objects must now indicate an internal status.


chris-grove:
I need to clean up a few things, so please wait on merging, apologies

markquintontulloch:
@oblodgett - AuditedObject class is defined in core.yaml
I'm not sure of the purpose of the table_key slot - I know it was in the previous mixin, but do we really want this?

model/schema/phenotypeAndDiseaseAnnotation.yaml
```diff
@@ -554,51 +552,51 @@ slots:
   has_condition:
     is_a: related_to
     symmetric: false
-    domain: Entity
+    domain: Association
```
markquintontulloch, line 555: Not sure what the rationale for this change is (and other changes in this file), but these slots are not used anywhere.  I think these are a hangover from before we decided to use Vocabulary and VocabularyTerm classes to deal with the controlled vocabularies and should be removed altogether - unless I am missing something.

chris-grove: Yep, good catch. I've just pushed a commit to delete those predicate definitions altogether

sierra-moxon (approved):

PR #131, opened 2022-05-10
SCRUM-1532 updating / adding / removing fields as appropriate to conf…

azurebrd:
…orm to new ABC database.  Fuller definitions for slots when data comes from pubmed, mod, or both

### DIFF
--- a/model/schema/reference.yaml
+++ b/model/schema/reference.yaml
@@ -40,10 +40,9 @@ classes:
     is_a: AuditedObject
     slots:
       - corresponding_author
+      - first_author
       - first_name
-      - middle_name
       - last_name
-      - initials
       - cross_references
 
   Reference:
@@ -54,18 +53,18 @@ classes:
       named references.
     slots:
       - abstract
+      - authors
       - category
-      - citation
       - curie
       - date_arrived_in_pubmed
+      - date_last_modified_in_pubmed
       - date_published
-      - issue_date
       - issue_name
       - keywords
       - language
       - merged_into_id
       - open_access
-      - pages
+      - page_range
       - plain_language_abstract
       - publisher
       - pubmed_abstract_languages
@@ -91,11 +90,6 @@ classes:
       reference_id:
         description: >-
           The primary key for the Reference object in the references table.
-      date_updated:
-        notes: >-
-          In the literature database, the values in this column actually
-          correspond to date_last_modified_in_pubmed, so the column should
-          be renamed in the literature database.
 
   MeshDetail:
     is_a: AuditedObject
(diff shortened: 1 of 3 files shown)


azurebrd:
Hi @sierra-moxon @gildossantos and @cerivs .  This is an attempt to resolve database changes at ABC to conform to LinkML.  It passes `make` so I hope it's okay.  
There's a couple of questions before merging (or can do another PR later).
https://www.wikidata.org/wiki/Property:P304 was describing `pages` which is now `page_range` because the data is a mix of PubMed and the MODs, which has pretty dirty data.  I think we should get rid of the  WIKIDATA_PROPERTY:P304  exact_mappings, but I don't know who added it, so I wanted to ask you.
Plain_language_abstract is not well defined, I've asked Ceri if she knows whether it's a layperson description or something else.
I don't think anything affects A-team, so I'm not tagging them.

model/schema/reference.yaml
```diff
@@ -199,122 +191,125 @@ slots:
...
     multivalued: true
     domain: Reference
     range: AuthorReference
 
   corresponding_author:
-    description: >-
-    range: Reference
+    description: Indicates if the author is a corresponding author.
+    range: boolean
     domain: Reference
```
gildossantos, line 282: 1. Should add `range: AuthorReference`?
2. Should add `multivalued: true`? Usually only one corresponding author, but are there edge cases where a paper has many corresponding authors?

model/schema/reference.yaml
```diff
@@ -199,122 +191,125 @@ slots:
...
 
-
-  initials:
-    description: >-
+  first_author:
+    description: Indicates if the author is a first author.
+    range: boolean
     domain: Reference
-    range: Reference
```
gildossantos, line 293: Add `range: AuthorReference`?

azurebrd:
Thanks @gildossantos , I'll update it to `close_mappings` .  Oh, not at all about `plain_language_abstract`, it's definitely correct to have, and we're getting that from PubMed, I just wanted to get a clear definition of it from Ceri and Kimberly, who tell me it's a layperson definition

model/schema/reference.yaml
gildossantos:
```suggestion
    domain: AuthorReference
```

model/schema/reference.yaml
```diff
@@ -199,122 +191,125 @@ slots:
...
+    range: boolean
     domain: Reference
 
-
-  initials:
-    description: >-
+  first_author:
+    description: Indicates if the author is a first author.
+    range: boolean
     domain: Reference
```
gildossantos:
```suggestion
    domain: AuthorReference
```

azurebrd:
@gildossantos Corresponding/First Authors :  Oh, yes AuthorReference, I'll update that.  I don't think it should be multivalued though, it's a boolean associated with each author object, instead of a list of properties associated to the paper.  I could be misunderstanding this though.  Yeah, apparently being a corresponding author or a first author is just a boolean associated with each author, and a given publication can have multiple of either.  (I'm not sure that it's 1+, it might be 0+)

gildossantos (changes requested):

model/schema/reference.yaml
gildossantos: Yes, I agree that those two slots (`first_author`, `corresponding_author`) should be `multivalued: false`. Sorry for the confusion. I had made comments to the contrary earlier, but deleted them almost immediately once I realized that those two slots have `domain: AuthorReference` rather than `domain: Reference`.

gildossantos (approved):

azurebrd:
Thanks so much @gildossantos !

PR #135, opened 2022-05-24
Gene yaml

gildossantos:


chris-grove (approved):
@gildossantos 
Thanks, it all looks good to me. I just merged #134 so please just make sure there are no conflicts with that.

PR #137, opened 2022-06-20
Proposal for dealing with allele (and other) evidence

markquintontulloch:
Here is my proposal for dealing with evidence for the allele class, although the same approach could be extended to other classes.  I believe this covers all WB's needs for the allele class, but other MODs may need additional changes.

I've gone for the approach of using specific intermediate classes to add evidence (and in some cases additional metadata, e.g. `origins`).  For example, the `generation_method` slot previously had a range of `string` whereas it now has a range of `GenerationMethod` - a class which contains the previous string and a slot pointing to an `Evidence` class.

The same thing could be achieved with associations, but I prefer this method for the following reasons:
- creates a more obvious link between the classes when browsing the model
- allows properties such as `required` and `multivalued` to be easily and obviously applied 
- I find the generated JSON schema easier to generate files for

The use of the `authors` slot in the `Evidence` class will need to be reviewed as I think the intent of the `authors` slot defined in `reference.yaml` is different to that of the WB `Author_evidence` tag (where Authors are people that cannot be unambiguously assigned to a WBPerson).

I think the only impact on the existing persistent store code is the addition of evidence to the Synonym class and the change of the way references are attached to the Note class.

Please let me know any thoughts you have on this.

### DIFF
--- a/jsonschema/allianceModel.schema.json
+++ b/jsonschema/allianceModel.schema.json
@@ -451,8 +451,8 @@
                "type": "string"
             },
             "database_status": {
-               "description": "Database status of the allele",
-               "type": "string"
+               "$ref": "#/$defs/DatabaseStatus",
+               "description": "Database status of the allele and associated evidence"
             },
             "date_created": {
                "description": "The date on which an entity was created. This can be applied to nodes or edges.",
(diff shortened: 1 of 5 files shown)


sbello:
Can we change generation method from a string to a simple vocabulary term? Having a pick list for these would be helpful for MGI folks. If it is a simple vocabulary we can just add all the terms groups are using without worrying about getting them in an ontology in the short term.

markquintontulloch:
@sbello - yes, I think vocabulary term would be a better fit for WB data too

sierra-moxon:
I'm just triaging our PR's in this repo, and I am going to close this for now in favor of our combined [PR](https://github.com/alliance-genome/agr_curation_schema/pull/152): please do reopen if I got this wrong.

PR #140, opened 2022-06-28
Change 'modified_by' to 'updated_by'

chris-grove:


gildossantos:
I pushed a change that updated the tests accordingly.

PR #144, opened 2022-07-11
biologicalEntitySet updated with draft WB fields

MagdalenaZZ:
Hi guys, 
I tried to add some slots to  GeneCluster,  GeneClass and  Operon, so that they will be able to harbour all WB data.
It is more of a draft than proper code, because I don't know exactly how to code it correctly, but I hope to learn :-)

### DIFF
--- a/model/schema/biologicalEntitySet.yaml
+++ b/model/schema/biologicalEntitySet.yaml
@@ -95,6 +95,7 @@ classes:
   FunctionalGeneSet:
     is_a: BiologicalEntity
     slots:
+      - genes
       - single_reference
 
   Pathway:
@@ -182,4 +183,11 @@ slots:
     domain: GeneNomenclatureSet
     range: Gene  
 
-      
+  experiment_type:
+    description: >-
+      Type of experiment by which these genes were collated eg chip-seq, interaction, expression
+    required: false
+    multivalued: true
+    domain: GeneCollection
+    range: string  # should maybe be replaced by a vocabulary
+
--- a/model/schema/biologicalEntitySet.yaml
+++ b/model/schema/biologicalEntitySet.yaml
@@ -196,4 +196,3 @@ slots:
     range: string  # should maybe be replaced by a vocabulary
 
 
-
--- a/model/schema/biologicalEntitySet.yaml
+++ b/model/schema/biologicalEntitySet.yaml
@@ -1,7 +1,7 @@
 id: https://github.com/alliance-genome/agr_curation_schema/biologicalEntitySet.yaml
 name: Alliance-Schema-Prototype-Biological-Entity-Set
 title: Alliance Schema Prototype Biological Entity Set
-description: Alliance Biological Entity Set Prototype with LinkML
+#description: Alliance Biological Entity Set Prototype with LinkML
 
 default_prefix: alliance
 default_range: string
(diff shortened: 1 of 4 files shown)


model/schema/biologicalEntitySet.yaml
```diff
@@ -56,18 +56,32 @@ classes:
     slots:
       - genes
 
+  GeneClass:
+    description: >-
+      A gene class is a set of genes which is of biological significance, and
```
sierra-moxon, line 61: Is it possible to describe this in more detail - what does biological significance mean?  And can you give an example of the kinds of phenotype attributed to a group of genes?

model/schema/biologicalEntitySet.yaml
```diff
@@ -56,18 +56,32 @@ classes:
     slots:
       - genes
 
+  GeneClass:
```
sierra-moxon, line 59: GeneClass vs. GeneSet vs. GeneGroup ;)  - is this a GeneFamily?

model/schema/gene.yaml
```diff
@@ -86,3 +94,31 @@ slots:
     description: SOTerm describing gene type
     domain: Gene
     range: SOTerm
+
+  gene_designating_laboratory:
```
markquintontulloch, line 98: I assume the intention of this slot is the same for Gene and GeneClass classes, so you can just use 'designating_laboratory' as defined in biologicalEntitySet.yaml and remove the `domain: GeneClass` line from the slot definition.  You can then either import biologicalEntitySet.yaml in gene.yaml or move the slot definition to core.yaml

model/schema/gene.yaml
```diff
@@ -86,3 +94,31 @@ slots:
...
+    multivalued: false
+    domain: GeneClass
+    range: Laboratory
+
+  designating_person:
+    description: >-
+      A person/persons who designated this gene
+    required: false
+    multivalued: true
+    domain: GeneClass
```
markquintontulloch, line 111: This should be Gene

model/schema/gene.yaml
```diff
@@ -86,3 +94,31 @@ slots:
...
+    range: Laboratory
+
+  splice_leader:
+    description: >-
+      Trans-splicing splice leaders observed to operate on this gene 
+    required: false
+    multivalued: true
+    domain: GeneClass
```
markquintontulloch, line 119: Also should be Gene

model/schema/gene.yaml
```diff
@@ -86,3 +94,31 @@ slots:
...
+
+  gene_designating_laboratory:
+    description: >-
+      A laboratory which designated this gene
+    required: false
+    multivalued: false
+    domain: GeneClass
+    range: Laboratory
+
+  designating_person:
```
markquintontulloch, line 106: For multivalue slots we tend to pluralise the slot name, i.e. designating_persons

MagdalenaZZ: done

model/schema/gene.yaml
MagdalenaZZ: done

model/schema/biologicalEntitySet.yaml
MagdalenaZZ: description improved

model/schema/gene.yaml
markquintontulloch: It's perfectly valid LinkML, but if the slots are intended to capture the same information then I don't see a need for duplication of code

model/schema/gene.yaml
```diff
@@ -86,3 +105,107 @@ slots:
...
+
+  wbprocess:
+    description: >-
+      WB specific. Eg WBbiopr:00000015 (Corpse engulfment) 
+    required: false
+    multivalued: true
+    domain: Gene
+    range: string  # this should be changed to a WBbiopr term
+
+  genetic_map_chromosome:
```
markquintontulloch, line 173: Might be worth considering creating a GeneticMapPosition class?

markquintontulloch: I'm not seeing any changes?

model/schema/gene.yaml
markquintontulloch: 'fixed' or 'decided against' - I don't see any changes?

model/schema/gene.yaml
markquintontulloch: Now I see them in the latest commit :)

MagdalenaZZ:
ok - seems like I could do it myself. Happy days :-)

PR #145, opened 2022-07-13
Added curie to DiseaseAnnotation and test examples

chris-grove:
Running 'make all' seems to work well. I get an error when running 'make test':
```
$ make test
rm -rf target/*
rm -rf jsonschema/*
mkdir -p target/jsonschema
poetry run gen-json-schema  --closed -t ingest model/schema/allianceModel.yaml > target/jsonschema/allianceModel.schema.json
WARNING:JsonSchemaGenerator:Unrecognized prefix: MSigDB
cp -pr target/jsonschema .
poetry run jsonschema -i test/data/allele_test.json jsonschema/allianceModel.schema.json
The currently activated Python version 3.7.4 is not supported by the project (^3.8).
Trying to find and use a compatible version. 
Using python3.8 (3.8.3)
Failed to parse 'jsonschema/allianceModel.schema.json': Expecting value: line 1 column 1 (char 0)
make: *** [validate-allele_test] Error 1
```
It seems like it was OK using Python 3.8 but then failed to parse the JSON schema file

### DIFF
--- a/.github/workflows/check-pull-request.yaml
+++ b/.github/workflows/check-pull-request.yaml
@@ -60,4 +60,3 @@ jobs:
       - name: Validate JSON Schema
         run: |
           make test-jsonschema
-          make test-jsonschema_invalid
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ t:
 echo:
 	echo $(patsubst %,gen-%,$(TGTS))
 
-test: all test-jsonschema test-jsonschema_invalid
+test: all test-jsonschema
 
 install:
 	poetry install
--- a/jsonschema/allianceModel.schema.json
+++ b/jsonschema/allianceModel.schema.json
@@ -9,12 +9,12 @@
                "type": "string"
             },
             "asserted_allele": {
-               "description": "The allele to which something is manually asserted to be associated.",
-               "type": "string"
+               "$ref": "#/$defs/Allele",
+               "description": "The allele to which something is manually asserted to be associated."
             },
             "asserted_gene": {
-               "description": "The gene to which something is manually asserted to be associated.",
-               "type": "string"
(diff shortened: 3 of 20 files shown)


chris-grove:
I think we need to leave this unmerged until we have instantiated curie values and/or I think we need to make an exception to requiring the curie for submitted disease annotations into the persistent store.

markquintontulloch:
> I think we need to leave this unmerged until we have instantiated curie values and/or I think we need to make an exception to requiring the curie for submitted disease annotations into the persistent store.

@chris-grove  I wonder if we need a disambiguation between external curies and AGR curies.  All curies currently submitted are external curies.  AGR curies are always going to have to be optional at the point of submission as DQMs are not going to be submitting them in most cases.  What is going to happen to curies currently submitted for genes, for example - will they continue to be submitted to the curie field or would we then need to create the equivalent of the mod_entity_id field for MOD curies and have the AGR curie in the curie slot?

I see two options:
1. create an (optional) agr_curie slot and leave everything else as it currently is
2. create mod_entity_id slots for all classes where there is a MOD-submitted curie and an AGR curie

The former requires less work and, to me, seems like a clearer indication of what is required in each field and we won't end up with a mixture of some curies  in the curie field being AGR curies and others being MOD curies.

sierra-moxon:
I fixed the remaining tests here and removed the "invalid" tests from PRs and the `make test` target.  While I think they are helpful to DQMs, I think to infrequent modelers, the (valid) invalid test output masks the failing tests.

sierra-moxon:
I'm going to merge this in, but it did require significant change to the tests, so on our next submission to the curation system, we'll need to remember to give DQMs a bit more time to update their scripts.

sierra-moxon (approved):

PR #146, opened 2022-07-13
Add children, parents, ancestors, and descendants to OntologyTerm

chris-grove:
I've labeled the slots "is_a_children", "is_a_parents", "is_a_ancestors", "is_a_descendants" at least as a placeholder. I think we do need to be specific about the relationship type.

### DIFF
--- a/model/schema/ingest.yaml
+++ b/model/schema/ingest.yaml
@@ -160,7 +160,7 @@ slots:
   ontology_closure_ingest_set:
     description: >-
     domain: Ingest
-    range: OntologyClosure
+    range: OntologyTermClosure
     multivalued: true
     mixins:
       - object_set
--- a/model/schema/ingest.yaml
+++ b/model/schema/ingest.yaml
@@ -108,6 +108,14 @@ slots:
     mixins:
       - object_set
 
+  ontology_closure_ingest_set:
+    description: >-
+    domain: Ingest
+    range: OntologyClosure
+    multivalued: true
+    mixins:
+      - object_set
+
   gene_ingest_set:
     description: >-
     domain: Ingest
@@ -139,4 +147,5 @@ classes:
       - allele_variant_association_ingest_set
       - agm_ingest_set
       - sqtr_ingest_set
+      - ontology_closure_ingest_set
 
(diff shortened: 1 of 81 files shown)


oblodgett (approved):

sierra-moxon (changes requested):
Will we have more slots in this class for every kind of closure?  part_of equivalents for example?  

What about a more generic data structure:

```
OntologyTermClosure:
   slots:
      - parent
      - child
      - relationship_type
 
slots:
   closure_type:
        range: enum_of_closures # ie: is_a, part_of, etc # depending on use cases, this could be straight from the RO, or an enumerated list of closure types.
```

oblodgett:
@sierra-moxon yes... but we let hibernate generate the join table, vs specifying it out here. So the class has the lists... however in the db there is only 2 tables to represent the closure for is_a.

oblodgett:
As stated in todays DQM meeting Parent - Child relationships for some mods are not considered as closure, but the ancestors - descendants are considered closure. In others both are considered closure. So here is what I propose


```
OntologyTerm:
   slots:
     - parents
     - children

OntologyTermClosure:
   slots:
      - ancestor
      - descendant 
      - distance_between # if we need it
      
OntologyTermRelationship:
   slots:
      - parent
      - child
      - relationship_type
 
slots:
   parents:
       range: OntologyTermRelationship
       multivalued: true
   children:
       range: OntologyTermRelationship
       multivalued: true
   ancestor:
       range: OntologyTerm
   descendant:
       range: OntologyTerm
   parent:
       range: OntologyTerm
   child:
       range: OntologyTerm
   relationship_type:
        range: ROTerm # We load the RO Ontology its a small ontology.
        
```

sierra-moxon:
This is purposefully failing for the moment - I'd like to talk about the test data and what we expect here.

model/schema/ingest.yaml
```diff
@@ -260,6 +268,7 @@ classes:
       - allele_origin_association_ingest_set
       - agm_ingest_set
       - sqtr_ingest_set
+      - ontology_closure_ingest_set
```
oblodgett, line 271: @chris-grove @sierra-moxon not sure we need an ingest for this considering the data is coming directly from owl files.

PR #147, opened 2022-07-21
Proposed changes for data loading strategy

markquintontulloch:
FOR DEMONSTRATION OF PRINCIPLE ONLY - NOT INTENDED FOR MERGING

Suggested changes to allele model to avoid circular dependencies when loading.  Example JSON submission included as test file.


sierra-moxon (approved):

sierra-moxon:
I folded this into the allele PR here: https://github.com/alliance-genome/agr_curation_schema/pull/152
ok to close this one?  (the only difference is really me removing the /bin directory).

PR #153, opened 2022-08-11
SCRUM-1855: represent a range of identifiers

abecerra:
Two classes representing AGR identifier and a consecutive range of identifiers. These will be used in MaTI and future MaTI API consumers (A and Blue teams)

### DIFF
--- a/model/schema/core.yaml
+++ b/model/schema/core.yaml
@@ -240,6 +240,18 @@ classes:
   Protein:
     is_a: GenomicEntity
 
+  Identifier:
+    slots:
+      - counter
+      - subdomain_code
+      - subdomain_name
+      - curie
+
+  IdentifiersRange:
+    slots:
+      - first
+      - last
+
 ## ------------
 ## SLOTS
 ## ------------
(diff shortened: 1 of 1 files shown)


sierra-moxon (approved):
@abecerra - I imagine this is the way we will assign AGRKB identifiers to an object?  Do you think we could add a bit to the description to point that out? 

Also, would it make sense to have a pointer to a named controlled vocabulary with the subdomain_codes already in use?   This might be helpful for example, if the BlueTeam wants a domain for a new AGRKB type, they would now which CV in the ATeam interface to go looking for (eg. CV.name = 'AGRKB_domain_vocabulary' or something).

abecerra:
> @abecerra - I imagine this is the way we will assign AGRKB identifiers to an object? Do you think we could add a bit to the description to point that out?

Thank you @sierra-moxon, I've updated the description 

> Also, would it make sense to have a pointer to a named controlled vocabulary with the subdomain_codes already in use? This might be helpful for example, if the BlueTeam wants a domain for a new AGRKB type, they would now which CV in the ATeam interface to go looking for (eg. CV.name = 'AGRKB_domain_vocabulary' or something).

At the moment all we have, behind the VPN, is:
https://alpha-mati.alliancegenome.org/api/subdomain
or
https://beta-mati.alliancegenome.org/api/subdomain
or
https://mati.alliancegenome.org/api/subdomain

returning the existing codes and subdomains (only one for now)

As we are closing to start minting IDs in the curation app, your suggestion can be discussed, also with the Blue team.

PR #164, opened 2022-10-06
raise error if validation fails

gildossantos:
I find it helpful in my pipelines to have validation script raise an error if validation fails, so suggesting it here in case others might find it helpful too (or at least innocuous).


sierra-moxon (approved):

gildossantos:
Thanks @sierra-moxon

PR #166, opened 2022-10-11
Synonyms as Associations

gildossantos:


sierra-moxon:
Thanks Gil!  looks good!  I have one small question: does this work with the "nomenclature_event" slot?  (I guess two small questions: does this mean that names and abbreviation/symbol have synonym entries even if they are the "current" name/abbreviation/symbol?)

gildossantos:
Thanks @sierra-moxon
As currently modeled, yes, there would be an entity-synonym relationship even for the current synonyms (how we do it at FlyBase). Does not need to be that way - could have a separate slot in entities for the current_symbol, current_full_name, for example.
I'm working on the nomenclature event as a separate annotation object. I think it should all coexist but I can add it to the PR (tomorrow realistically) so it can be reviewed together.

markquintontulloch (commented):
Looks good, we definitely need consistency for synonyms.  I just a couple of thoughts / questions.

Are the relationships between entities and synonyms always one-to-many, or can they be many-to-many.   If the former, I wonder if it would be simpler to move the evidence and synonym_scope slots from the EntitySynonym class to the Synonym class itself and have a synonyms slot on the entity classes.

For DiseaseAnnotations we have avoided enums in the favour of VocabularyTerm objects to avoid having to hard code values and to enable curators to update the lists when required via the curation interface.   Would that also make sense for synonym_type, synonym_scope, and entity_synonym_type?

markquintontulloch:
yeah, definitely good to keep a record of valid terms in the LinkML model so that we know what terms to load into the persistent store when it comes to implementation - I guess these could be added to the description so that the range can be set as VocabularyTerm and we don't have to change the model when we implement, but maybe that's a bit clunky?

gildossantos:
Seems like a good idea to have those terms in the description of an object, at least until it gets implemented in the persistent. Upon implementation, we may still probably want to remove those terms from the description as we don't want to worry about keeping the terms in the LinkML in synch with the actual VocabularyTerms (source of truth).

PR #169, opened 2022-10-12
Synonym as SlotAnnotation (subject part of SlotAnnotation)

gildossantos:


sierra-moxon (approved):

model/schema/controlledVocabulary.yaml
```diff
@@ -80,3 +73,11 @@ classes:
...
         required: true
         identifier: true
+
+  VocabularyTermSynonym:
+    is_a: Synonym
+    description: >-
+      The relationship between a controlled vocabulary term and a synonym.
+    slot_usage:
+      subject:
+        range: VocabularyTerm
```
markquintontulloch, line 83: Is this for the synonym of a VocabularyTerm?  If so, VocabularyTerm is the domain, not the range.  Also, I don't see a slot on VocabularyTerm pointing to this.

markquintontulloch, line 83: The VocabularyTerm is the domain, not the range - same comment goes for OntologySynonym and ResourceSynonym.  Also, I don't see any slots linking these classes to the VocabularyTerm / Ontology / Resource.

Do you envisage that there will eventually be different slots for the synonyms of each of these classes, or could we just use a multivalued 'synonyms' slot that can be added to all these classes and has a range of Synonym?

model/schema/ontologyTerm.yaml
```diff
@@ -161,6 +156,14 @@ classes:
           distance_between is zero for reflexive–transitive closure
           each node has an ancestor or descendant of itself 
 
+  OntologyTermSynonym:
+    is_a: Synonym
+    description: >-
+      The relationship between an ontology term and a synonym.
+    slot_usage:
+      subject:
+        range: OntologyTerm
```
markquintontulloch, line 165: see above comment

model/schema/resource.yaml
```diff
@@ -70,6 +69,14 @@ classes:
         description: >-
           the title of the publication
 
+  ResourceSynonym:
+    is_a: Synonym
+    description: >-
+      The relationship between a resource and a synonym.
+    slot_usage:
+      subject:
+        range: Resource
```
markquintontulloch, line 78: see above comment

markquintontulloch (changes requested):

model/schema/controlledVocabulary.yaml
markquintontulloch: duplicate comment, sorry

gildossantos: Yes, this is a synonym for a vocabulary term. The proposed VocabularyTerm entity has a subject slot that points to the VocabularyTerm, rather than a synonyms slot in the VocabularyTerm that points to a Synonym.

gildossantos: Maybe I'm confused, but I've looked at this again, and as written, I think the subject in these cases is correct. For example, we have a VocabularyTermSynonym object (not a slot of a VocabularyTerm, but its own object), and the subject slot for that object (to which the synonym applies) must refer to a VocabularyTerm.

Also, because these Synonym objects (AlleleSynonym, ResourceSynonym, etc) already have a subject slot that points to the relevant entity (e.g., gene, allele, resource, etc), I had not envisioned a need for a synonyms slots in the entities that have synonyms. 

I'm guessing that my proposal is somewhat different from what you had in mind? I chose this path because there are many-to-many entity-synonym relationships, and it seemed better to specify the entity (e.g., vocabulary term) to which the synonym applies right here in the Synonym entity.

gildossantos:
So I think the suggestion from @markquintontulloch is to remove the `subject` slot from the proposed `Synonym` object, and replace that with a `synonyms` slot in the entity that has synonyms. I will come up with a 3rd proposal in line with that approach.

My hesitation is that because there are many-to-many relationships (e.g., a given synonym can refer to one gene in one set of papers, and a different gene in a different papers), we'll have cases where synonym entities are the same except for the evidence associated, and it's not clear to me how those similar synonyms are clearly assigned to the correct entities. I suppose it should be clear in an ingest file, and perhaps that's all that matters. If I could understand how that gets represented in postgres (perhaps a new join table), then I would feel more comfortable going this route.

gildossantos:
I did make the Synonym `is_a: SlotAnnotation` (so as not to create a new design pattern), though I ended up with something maybe more complicated than envisioned for the parent class.

It might worth discussing the alternatives at an upcoming DQM meeting before deciding which one to go with.

markquintontulloch:
We already have many-to-many relationships directly attached to DiseaseAnnotations (e.g. ConditionRelations).  In that particular case we generate and store uniqueId by concatenating fields in order to be able to assign ConditionRelations to DiseaseAnnotations, but this isn't necessary as we can just compare the fields that define uniqueness.

PR #170, opened 2022-10-12
Synonyms as SlotAnnotation (subject linked via its own synonyms slot)

gildossantos:


markquintontulloch:
This is my favourite of the proposals, so long as it meets all your use cases.

PR #171, opened 2022-10-12
Draft EventHistory (nomenclature tracking)

gildossantos:


MagdalenaZZ (approved):
Could we think of re-naming EventHistory to EventHistoryEntry or EventEntry or something like that, to reflect that the object may not be the entire history, but just one event?

gildossantos:
Yes, good point - I've changed the name to `EventHistoryEntry`.

PR #173, opened 2022-10-25
Proposal for separation of ingest and database concerns in LinkML model

markquintontulloch:
As a result of a discussion within the A-Team, we believe the best way forward or the LinkML model is to separate ingest modelling from database / Java code modelling.  This is an initial proposed remodel for the classes currently ingested by the A-Team's curation database.  This accurately represents both our ingest DTO classes and our entity classes (used to generate the database schema via Hibernate).  It allows for fields to be required in the database but not in the ingest files, and allows both forward- and back-references to be explicitly described in the LinkML model.  

@oblodgett, @chris-grove  - Any initial feedback before taking this to the wider modelling group?

### DIFF
--- a/model/schema/affectedGenomicModel.yaml
+++ b/model/schema/affectedGenomicModel.yaml
@@ -47,13 +47,30 @@ classes:
       - data_provider
       - references
 
+  AffectedGenomicModelDTO:
+    is_a: GenomicEntityDTO
+    description: Ingest class for AGMs
+    slots:
+      - subtype_name
+      - reference_curies
+      - data_provider
+      - reference_curies
+      - sequence_targeting_reagent_curies
+      - component_dtos
+
   AffectedGenomicModelComponent:
     is_a: AuditedObject
     description: >-
       Allele that affects the model and its zygosity
     slots:
-      - has_allele
+      - single_allele
       - zygosity
+
+  AffectedGenomicModelComponentDTO:
+    is_a: AuditedObjectDTO
+    slots:
+      - allele_curie
+      - zygosity_curie
 
 slots:
 
@@ -63,10 +80,18 @@ slots:
 
   subtype:
     description: >-
-      Subtype of affected genomic model
+      Subtype of affected genomic model - permissible values: strain / genotype
+      / fish
     domain: AffectedGenomicModel
-    range: subtype_values
+    range: VocabularyTerm
     required: true
+
+  subtype_name:
+    description: >-
+      Name of VocabularyTerm describing subtype - permissible values: strain / genotype
+      / fish
+    domain: AffectedGenomicModelDTO
+    range: string
 
   components:
     singular_name: component
@@ -75,6 +100,11 @@ slots:
       i.e. 'allele', each with a zygosity
     domain: AffectedGenomicModel
     range: AffectedGenomicModelComponent
+    multivalued: true
+
+  component_dtos:
+    domain: AffectedGenomicModelDTO
+    range: AffectedGenomicModelComponentDTO
     multivalued: true
 
   has_allele:
@@ -87,10 +117,24 @@ slots:
     domain: AffectedGenomicModelComponent
     range: zygosity_values
 
+  zygosity_curie:
+    description: >-
+      Curie of GENO ontology ID for allele zygosity - permissible_values:
+      GENO:0000602 / GENO:0000603 / GENO:0000604 / GENO:0000605
+      / GENO:0000606 / GENO:0000135 / GENO:0000136 / GENO:0000137
+      / GENO:0000134
+    domain: AffectedGenomicModelComponentDTO
+    range: string
+
   sequence_targeting_reagents:
     singular_name: sequence_targeting_reagent
     domain: AffectedGenomicModel
     range: SequenceTargetingReagent
+    multivalued: true
+
+  sequence_targeting_reagent_curies:
+    domain: AffectedGenomicModelDTO
+    range: string
     multivalued: true
 
   parental_populations:
--- a/model/schema/affectedGenomicModel.yaml
+++ b/model/schema/affectedGenomicModel.yaml
@@ -57,6 +57,10 @@ classes:
 
 slots:
 
+  agm_curie:
+    range: string
+    required: true
+
   subtype:
     description: >-
       Subtype of affected genomic model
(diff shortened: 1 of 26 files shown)


sierra-moxon:
I think this is excellent!!

model/schema/core.yaml
```diff
@@ -156,28 +202,40 @@ classes:
...
+      - free_text
+      - note_type_name
+      - evidence_curies
+    slot_usage:
+      free_text:
+        required: true
+      note_type_name:
+        required_true
+
   SlotAnnotation:
```
sierra-moxon, line 217: remove this class?

markquintontulloch: Yeah, I think there's probably a bunch of cleanup to be done before moving this PR from draft - just wanted to get feedback first

markquintontulloch:
@sierra-moxon @chris-grove The proposal has now been updated to use SlotAnnotations again (with one less stage of inheritance).  For the slot names representing the 'backlinks' I have kept the names the same rather than adding '_slot_annotation', e.g. 'allele_mutation_types' vs. 'allele_mutation_type_slot_annotations' - changing them seemed excessively verbose but maybe this would be preferred for the sake of clarity (I don't feel particularly strongly either way).  I did change the SlotAnnotation class names though just to make the design pattern a bit more obvious without relying on noticing the `is_a: SlotAnnotation`.

model/schema/affectedGenomicModel.yaml
```diff
@@ -47,23 +47,52 @@ classes:
...
       - references
 
+  AffectedGenomicModelDTO:
+    is_a: GenomicEntityDTO
+    description: Ingest class for AGMs
+    slots:
+      - subtype_name
+      - reference_curies
+      - data_provider
+      - reference_curies
```
chris-grove, line 57: slot 'reference_curies' present twice

model/schema/allele.yaml
```diff
@@ -54,6 +54,23 @@ classes:
...
+      - allele_notes
+      - allele_synonyms
+      - allele_gene_associations
+      - allele_transcript_associations
+      - allele_protein_associations
+      - allele_variant_associations
+      - allele_construct_associations
+      - allele_cell_line_associations
+      - allele_image_associations
+      - allele_origin_associations
```
chris-grove, line 73: missing 'allele_generation_method_associations' slot

model/schema/core.yaml
```diff
@@ -156,28 +201,47 @@ classes:
...
+    is_a: AuditedObjectDTO
+    slots:
+      - free_text
+      - note_type_name
+      - evidence_curies
+    slot_usage:
+      free_text:
+        required: true
+      note_type_name:
+        required_true
```
chris-grove, line 214: required: true

markquintontulloch:
@chris-grove Good spots!  I've fixed the above comments, updated the ingest sets with a couple of missing allele associations, and also removed sequencing_status from the allele model.  Do you want me to change inheritance_mode to be a SlotAnnotation in this PR, or shall we do that as a separate PR?

model/schema/allele.yaml
```diff
@@ -207,6 +356,13 @@ classes:
       object:
         range: CellLine
 
+  AlleleCellLineAssociationDTO:
+      is_a: AuditedObjectDTO
+      slots:
+        - allele_curie
+        - predicate_name
+        - cell_line_curie
+
```
chris-grove, line 365: because this is not inheriting from an 'evidence' slot-bearing class (like a hypothetical AssociationDTO class), 'evidence_curies' needs to be added here

model/schema/allele.yaml
```diff
@@ -225,15 +381,13 @@ classes:
...
-        range: Allele
-      object:
-        range: Note
+  AlleleImageAssociationDTO:
+    is_a: AuditedObjectDTO
+    slots:
+      - allele_curie
+      - predicate_name
+      - image_curie
+      - primary_image
```
chris-grove, line 390: because this is not inheriting from an 'evidence' slot-bearing class (like a hypothetical AssociationDTO class), 'evidence_curies' needs to be added here

model/schema/allele.yaml
```diff
@@ -245,18 +399,12 @@ classes:
...
-      - synonym_scope
-    slot_usage:
-      subject:
-        range: Allele
-      object:
-        range: Synonym
-
+      - allele_curie
+      - predicate_name
+      - agm_curie
```
chris-grove, line 407: because this is not inheriting from an 'evidence' slot-bearing class (like a hypothetical AssociationDTO class), 'evidence_curies' needs to be added here

markquintontulloch:
@chris-grove - evidence_curies slots added.  We probably want to revise the Association and AssociationDTO classes at some point anyway to ensure the appropriate ranges are being used for predicates and to give them more descriptive names in the DTO classes.  Also AlleleGeneAssociation is currently inheriting an ROTerm predicate, which I think we decided yesterday was incorrect.

chris-grove:
@markquintontulloch OK sounds good! Yes, the ROTerm needs to be changed as the inherited predicate range (probably to VocabularyTerm) and the inheritance_mode needs to be a SlotAnnotation, but I/we can deal with that in our next iteration.

model/schema/allele.yaml
```diff
@@ -225,15 +382,14 @@ classes:
...
-        range: Allele
-      object:
-        range: Note
+  AlleleImageAssociationDTO:
+    is_a: AuditedObjectDTO
+    slots:
+      - allele_curie
+      - predicate_name
+      - image_curie
+      - primary_image
```
chris-grove, line 391: 'primary_image' slot range is not specified; seems that it should be a boolean

model/schema/allele.yaml
```diff
@@ -324,12 +482,25 @@ slots:
     description: >-
```
chris-grove, line 482: 'primary_image' slot range is not specified; seems that it should be a boolean

markquintontulloch: Boolean seems sensible, but it’s just carried over from the current repo - not a new slot - so I’m not sure.  There’s definitely tidying up to do on the allele model but the aim of this PR was simply to transition to a model with the new paradigm of separate ingest classes.

markquintontulloch: Are you happy enough with the new parts of the model that we can merge @chris-grove ?

chris-grove (approved):
Whew! That's quite the PR! I've gone through everything now and I think everything should be good.

PR #174, opened 2022-10-27
SCRUM-2182 VocabularyTermSet

markquintontulloch:


model/schema/controlledVocabulary.yaml
```diff
@@ -80,3 +94,18 @@ classes:
...
+
+  VocabularyTermSet:
+    is_a: AuditedObject
+    description: >-
+      A subset of terms from a Vocabulary that are valid for particular
+      applications
+    slots:
+      - name
+      - set_vocabulary
+      - set_description
```
oblodgett, line 106: I know its a long name but would it be better to use vocabulary_term_set_description, in the future we might want to have generic "sets" of things.

markquintontulloch: sure - should we do the same for set_vocabulary -> vocabulary_term_set_vocabulary ?

markquintontulloch: @oblodgett - done, I changed set_vocabulary too

oblodgett (approved):

PR #179, opened 2022-11-14
NameSlotAnnotation - more granular

gildossantos:


model/schema/allele.yaml
```diff
@@ -61,7 +61,7 @@ classes:
       - allele_secondary_ids
       - allele_nomenclature_events
       - allele_notes
-      - allele_synonyms
+      # - allele_synonyms    # Allele inherits synonyms slot from GenomicEntity
```
markquintontulloch, line 64: From the Java perspective I think we would need to keep the `allele_synonyms` slot (but change the range to `AlleleSynonymSlotAnnotation`) and remove the `synonyms` slot from `BiologicalEntity`, rather than specifying the range in the slot_usage.  This is because we would need to specify the 'back reference' to the AlleleSynonymSlotAnnotation table in the Allele class.

gildossantos: Thanks Mark. I'll restore `allele_synonyms` and remove the generic `synonyms` slot. It seems like we should also have class specific slots like `allele_symbol` or `gene_full_name` rather than generic `symbol` and `full_name` slots in the `core.yaml`. Thoughts?

model/schema/gene.yaml
```diff
@@ -111,14 +143,48 @@ classes:
...
 
-
 ## ------------
 ## SLOTS
 ## ------------
 
+slots:
+
+  single_gene:
+    domain: Gene
```
chris-grove, line 153: this should be "range: Gene" right? with no specification of domain?

gildossantos: Good catch. Fixed.

gildossantos:
@sierra-moxon - could you have a look at my attempt to constrain `name_type` in various `NameSlotAnnotation` objects like the `AlleleSymbolSlotAnnotation` and make sure it looks ok?
@chris-grove - I added descriptions to the various `NameSlotAnnotation` objects.

sierra-moxon (approved):
looks great!  thanks @gildossantos

markquintontulloch:
@gildossantos DTO classes and slots now added and test files updated

gildossantos:
thanks @markquintontulloch !

model/schema/allele.yaml
```diff
@@ -831,15 +854,51 @@ slots:
...
+    required: false
+    multivalued: false
+
+  allele_full_name_dto:
+    description: >-
+      The one current full name for an allele: e.g., wg<sup>1</sup>.
+    domain: AlleleDTO
+    range: FullNameSlotAnnotationDTO
+    multivalued: false
+    required: true
```
chris-grove, line 890: I believe this shouldn't be required, only the allele_symbol_dto

markquintontulloch: Good spot - I've removed the requirement

model/schema/core.yaml
```diff
@@ -376,13 +394,65 @@ classes:
...
+        retired_name: SGD-specific.
+        unspecified: unclassified names.
+    domain: NameSlotAnnotation
+    range: VocabularyTerm
+    multivalued: false
+    required: true
+
+  name_type_name:
+    description: >-
+      Name of the VocabularyTerm representing the name tpye of the
```
chris-grove, line 417: type

markquintontulloch: typo now corrected

model/schema/core.yaml
```diff
@@ -376,13 +394,65 @@ classes:
...
+    range: VocabularyTerm
+    multivalued: false
+    required: true
+
+  name_type_name:
+    description: >-
+      Name of the VocabularyTerm representing the name type of the
+      synonym - proposed values are nomenclature_symbol / full_name /
+      systematic_name / ncbi_protein_name / uniform / non_uniform /
+      retired_name / unspecified
```
chris-grove, line 420: Probably worth declaring the domain as "NameSlotAnnotationDTO" as it isn't used elsewhere and probably shouldn't be? I do see domains specified for other DTO class slots

model/schema/core.yaml
```diff
@@ -376,13 +394,65 @@ classes:
...
+    range: uri
 
   synonym_scope:
     description: >-
       the scope of the synonym - permissible values are narrow /
       broad / related / exact
+    domain: NameSlotAnnotation
     range: VocabularyTerm
 
   synonym_scope_name:
```
chris-grove, line 458: Probably worth declaring the domain as "NameSlotAnnotationDTO" as it isn't used elsewhere and probably shouldn't be? I do see domains specified for other DTO class slots (see also slot 'name_type_name')

model/schema/core.yaml
```diff
@@ -376,13 +394,65 @@ classes:
...
+    range: string
+    required: true
+
+  format_text:
+    description: >-
+      A version of a synonym string using only ASCII characters, which is
+      easier to type (for searches), print and parse. For example, Greek
+      characters are transliterated.
+    aliases: ['synonym_text']
+    domain: NameSlotAnnotation
```
chris-grove, line 430: This domain constraint should be removed (or at least relaxed/expanded, if possible) as this slot is also used by the NameSlotAnnotationDTO class

model/schema/core.yaml
```diff
@@ -376,13 +394,65 @@ classes:
...
+    range: string
+    required: true
+    multivalued: false
+
+  display_text:
+    description: >-
+      A version of a synonym string for display. Any UTF8 character is
+      permitted.
+    aliases: ['synonym_sgml']
+    domain: NameSlotAnnotation
```
chris-grove, line 440: This domain constraint should be removed (or at least relaxed/expanded, if possible) as this slot is also used by the NameSlotAnnotationDTO class

model/schema/core.yaml
```diff
@@ -376,13 +394,65 @@ classes:
...
+    range: string
+    required: true
+    multivalued: false
+
+  synonym_url:
+    description: >-
+      URL for a synonym: e.g., NCBI URL for the NCBI synonym of an object.
+    domain: NameSlotAnnotation
```
chris-grove, line 448: This domain constraint should be removed (or at least relaxed/expanded, if possible) as this slot is also used by the NameSlotAnnotationDTO class

model/schema/core.yaml
markquintontulloch: This one should stay as it is.  `synonym_scope` is for the non-DTO class.

model/schema/core.yaml
markquintontulloch: Done

model/schema/core.yaml
markquintontulloch: Done

model/schema/core.yaml
markquintontulloch: Done

model/schema/core.yaml
markquintontulloch: Done

model/schema/allele.yaml
```diff
@@ -235,25 +233,41 @@ classes:
...
       - single_allele
-      - synonym
-      - synonym_scope
     slot_usage:
       single_allele:
         required: true
-      synonym:
+      name_type:
+        any_of:
+          - equals_string: symbol
```
chris-grove, line 247: I'm wondering: I know the name_type_name is a DTO class slot and so the range is "string" and therefore we can use the "equals_string" assertion, but I wonder if we can apply the same constraint for "name_type" where the range is actually 'VocabularyTerm'?

model/schema/allele.yaml
```diff
@@ -235,25 +233,41 @@ classes:
...
+      The one current full name for the allele.
+    is_a: NameSlotAnnotation
+    slots:
+      - single_allele
+    slot_usage:
+      single_allele:
         required: true
+      name_type:
+        any_of:
+          - equals_string: full_name
```
chris-grove, line 261: I'm wondering: I know the name_type_name is a DTO class slot and so the range is "string" and therefore we can use the "equals_string" assertion, but I wonder if we can apply the same constraint for "name_type" where the range is actually 'VocabularyTerm'?

model/schema/gene.yaml
```diff
@@ -85,9 +86,61 @@ classes:
...
+      The one current symbol for the gene.
+    is_a: NameSlotAnnotation
+    slots:
+      - single_gene
+    slot_usage:
+      single_gene:
+        required: true
+      name_type:
+        any_of:
+          - equals_string: nomenclature_synonym
```
chris-grove, line 106: Like last two comments: I'm wondering: I know the name_type_name is a DTO class slot and so the range is "string" and therefore we can use the "equals_string" assertion, but I wonder if we can apply the same constraint for "name_type" where the range is actually 'VocabularyTerm'?

model/schema/gene.yaml
```diff
@@ -85,9 +86,61 @@ classes:
...
+    description: >-
+      The one current systematic name for the gene.
+    slots:
+      - single_gene
+    slot_usage:
+      single_gene:
+        required: true
+      name_type:
+        any_of:
+          - equals_string: systematic_name
```
chris-grove, line 132: Like last several comments: I'm wondering: I know the name_type_name is a DTO class slot and so the range is "string" and therefore we can use the "equals_string" assertion, but I wonder if we can apply the same constraint for "name_type" where the range is actually 'VocabularyTerm'?

model/schema/allele.yaml
markquintontulloch: I had the same thought, but I'm not sure these "rules" have much impact on anything (at least not anything we are using LinkML for) so I left them in as a potentially useful reminder for anyone looking at the model and not looking at the DTO classes.  Maybe better to add them as `notes` under `slot_usage` - I'll make that change

model/schema/allele.yaml
markquintontulloch: Changed to note

model/schema/allele.yaml
markquintontulloch: Changed to note

model/schema/gene.yaml
```diff
@@ -85,9 +86,61 @@ classes:
...
+    description: >-
+      The one current full name for the gene.
+    slots:
+      - single_gene
+    slot_usage:
+      single_gene:
+        required: true
+      name_type:
+        any_of:
+          - equals_string: full_name
```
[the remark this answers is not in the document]
markquintontulloch: Changed to note

model/schema/gene.yaml
markquintontulloch: Changed to note

model/schema/gene.yaml
```diff
@@ -102,8 +102,8 @@ classes:
       single_gene:
         required: true
       name_type:
-        any_of:
-          - equals_string: nomenclature_synonym
+        name_type:
+          notes: value must be VocabularyTerm with name "symbol"
```
markquintontulloch, line 106: @gildossantos - Is this correct?  The `any_of` rule previously stated it should be `nomenclature_synonym`, but the  `AlleleSymbolSlotAnnotation` states it should be `symbol`

model/schema/gene.yaml
markquintontulloch: Change to note

chris-grove:
@markquintontulloch 
I see we have DTO classes:
```
SymbolSlotAnnotationDTO
FullNameSlotAnnotationDTO
SystematicNameSlotAnnotationDTO
```
as children of 'NameSlotAnnotationDTO' and which are the ranges for corresponding slots:
```
allele_symbol_dto / gene_symbol_dto
allele_full_name_dto / gene_full_name_dto
gene_systematic_name_dto
```
wheras the synonyms slots have ranges directly pointing to 'NameSlotAnnotationDTO':
```
  allele_synonym_dtos:
    domain: AlleleDTO
    range: NameSlotAnnotationDTO
    multivalued: true
    inlined: true
    inlined_as_list: true
```
Do we not also need the DTO class for synonyms? Like:
```
  SynonymSlotAnnotationDTO:
    is_a: NameSlotAnnotationDTO
    slot_usage:
      name_type_name:
        any_of:
          - equals_string: synonym
```
for the same reason as the others?

markquintontulloch:
@chris-grove No, we have specific DTO classes for SymbolSlotAnnotation, FullNameSlotAnnotation, and SystematicNameSlotAnnotation in order to be able to specify the class-specific rules regarding name_type_name.  This is not necessary for the other name types.

test/data/allele_slot_annotation_ingest_test.json
```diff
@@ -79,6 +99,19 @@
           ],
           "internal": false
         }
+      ],
+      "allele_synonym_dtos": [
+        {
+          "name_type_name": "systematic_name",
```
chris-grove, line 105: should be "synonym" no?

markquintontulloch: My understanding was that synonym could be any type of name.

chris-grove:
@markquintontulloch @gildossantos 
OK, I'm all set with the pull request if we resolve my last comment:
https://github.com/alliance-genome/agr_curation_schema/pull/179#discussion_r1032545947
Not a big deal, but should the "name_type_name" for the 'allele_synonym_dtos' slot in the test file 'allele_slot_annotation_ingest_test.json' be 'synonym' rather than 'systematic_name'?

test/data/allele_slot_annotation_ingest_test.json
markquintontulloch: There are no rules in place for name_type for `AlleleSymbolSlotAnnotation` (or the DTO equivalents), and IMO 'systematic_name' would be a valid name type - in the case of Alleles there is no other slot for systematic names (if anyone has them), and in the case of genes you may want to add a systematic name that is not the one current systematic name that would go under `GeneSystematicNameSlotAnnotation`.

Also, based on the `name_type` slot definition in core.yaml, I don't believe that 'synonym' is one of the proposed name types (although based on the list in the description there is a discrepancy between 'nomenclature_symbol' used in the slot description and 'symbol' used in the rules/notes for the SymbolSlotAnnotation classes.

chris-grove (approved):

test/data/allele_slot_annotation_ingest_test.json
gildossantos: @chris-grove - yes, synonym can be of any type. Can have old "full_name" or "systematic_name" synonyms in here.
@markquintontulloch - I agree that "systematic_name" can be valid for symbol (for cases where gene or allele has not yet been named):e.g., FB has systematic names for genes and alleles where no symbol exists. I've updated the `SymbolSlotAnnotationDTO`, `AlleleSymbolSlotAnnotation` and `GeneSymbolSlotAnnotation` objects accordingly.

model/schema/gene.yaml
gildossantos: @markquintontulloch - sorry, that was a typo. It should be "nomenclature_symbol". I've updated the `SymbolSlotAnnotationDTO`, `AlleleSymbolSlotAnnotation` and `GeneSymbolSlotAnnotation` objects accordingly.

gildossantos:
I've made some minor corrections and clarifications.

It should "nomenclature_symbol" synonym type for symbols.
https://github.com/alliance-genome/agr_curation_schema/pull/179#discussion_r1033638980

Allow "systematic_name" type for symbol, clarify that synonym can be of any type.
https://github.com/alliance-genome/agr_curation_schema/pull/179#discussion_r1033637064

I have one question about the DTOs. We've defined various generic types of name slot annotation DTOs in core, but the non-DTO version are only declared in the gene and allele yaml files. To be consistent, should we also make generic objects like `SymbolSlotAnnotation` in core.yaml to mirror the `SymbolSlotAnnotationDTO`, etc?

chris-grove:
> I've made some minor corrections and clarifications.
> 
> It should "nomenclature_symbol" synonym type for symbols. [#179 (comment)](https://github.com/alliance-genome/agr_curation_schema/pull/179#discussion_r1033638980)
> 
> Allow "systematic_name" type for symbol, clarify that synonym can be of any type. [#179 (comment)](https://github.com/alliance-genome/agr_curation_schema/pull/179#discussion_r1033637064)
> 
> I have one question about the DTOs. We've defined various generic types of name slot annotation DTOs in core, but the non-DTO version are only declared in the gene and allele yaml files. To be consistent, should we also make generic objects like `SymbolSlotAnnotation` in core.yaml to mirror the `SymbolSlotAnnotationDTO`, etc?

@gildossantos 
What had been agreed on earlier was that we would skip instantiation of a generic "SymbolSlotAnnotation" class to go directly from "SlotAnnotation" to the child class "<Class>SymbolSlotAnnotation" like "AlleleSymbolSlotAnnotation" or "GeneSymbolSlotAnnotation", and as these are class-specific it makes sense to keep them in the class-specific yaml files and not core. @markquintontulloch @sierra-moxon can correct me if I misunderstood.

gildossantos:
@chris-grove - that's my understanding too, but why doesn't that apply to the `NameSlotAnnotationDTO` objects then?

markquintontulloch:
@gildossantos - we don't have class-specific versions of the`SymbolSlotAnnotationDTO`, `FullNameSlotAnnotationDTO`, or `SystematicNameSlotAnnotationDTO` classes as, because the DTO classes are inlined, we don't have the gene/allele/whatever submitted as part of the slot annotation.

gildossantos:
Thanks for explaining @markquintontulloch.

PR #182, opened 2022-12-04
KANBAN-265 Updating Allele model for WB Transgenes

chris-grove:
Summary of changes:
allele.yaml
- Add 'agent' and 'phenotypeAndDiseaseAnnotation' imports
- Add 'laboratory_of_origin', 'is_extrachromosomal', 'is_integrated', 'allele_inheritance_modes', 'allele_allele_associations' and 'allele_genomic_location_associations' slots to Allele class model
- Add 'laboratory_of_origin_curie', 'is_extrachromosomal', 'is_integrated', and 'allele_inheritance_mode_dtos' slots to AlleleDTO class model
- Add 'integration_method', 'chemical_mutagen', and 'irradiation_mutagen' slots to GenerationMethod class
- Add 'integration_method_name', 'chemical_mutagen_name', and 'irradiation_mutagen_name' slots to GenerationMethodDTO class
- Add 'AlleleInheritanceModeSlotAnnotation' and 'AlleleInheritanceModeSlotAnnotationDTO' Slot Annotation classes including:
  - 'inheritance_mode', 'phenotype_term', and 'phenotype_statement' slots to AlleleInheritanceModeSlotAnnotation
  - 'inheritance_mode_name', 'phenotype_term_curie', and 'phenotype_statement' slots to AlleleInheritanceModeSlotAnnotationDTO
- Add 'phenotype_term' and 'phenotype_statement' slots to AlleleFunctionalImpactSlotAnnotation
- Add 'phenotype_term_curie' and 'phenotype_statement' slots to AlleleFunctionalImpactSlotAnnotationDTO
- AlleleGenomicEntityAssociation class
  * Removed 'evidence' slot (redundant with Association class)
  * Changed predicate range from ROTerm to VocabularyTerm
  * Removed evidence requirement
- Add AlleleAlleleAssociation class
- Add predicate range of VocabularyTerm (with note) to AlleleConstructAssociation class
- Add slot definitions for slots: 'laboratory_of_origin', 'laboratory_of_origin_curie', 'is_extrachromosomal', 'is_integrated', 'allele_inheritance_modes', 'allele_inheritance_mode_dtos', 'allele_allele_associations', 'integration_method', 'integration_method_name', 'chemical_mutagen', 'chemical_mutagen_name', 'irradiation_mutagen', and 'irradiation_mutagen_name'
- Changed GenomicLocation to GenomicLocationAssociation and added appropriate DTO class and slots

phenotypeAndDiseaseAnnotation.yaml
- Add slot definition for 'phenotype_statement' and 'phenotype_term_curie'

ingest.yaml
- Added 'allele_genomic_location_association_ingest_set'

test files
- Update 'symbol' name_type_name to 'nomenclature_symbol'
- Add example WB transgene
- Update 'allele_inheritance_mode_names' to 'allele_inheritance_mode_dtos'
- Added 'allele_genomic_location_association_ingest_set'


model/schema/allele.yaml
```diff
@@ -464,6 +560,14 @@ classes:
       - agm_curie
       - evidence_curies
 
+
+  AlleleGenomicLocationAssociation:
+    is_a: GenomicLocationAssociation
```
markquintontulloch, line 565: `GenomicLocationAssociation` doesn't exist.  I think this should just be '`is_a: Association`.

markquintontulloch (changes requested):

test/data/allele_test.json
```diff
@@ -58,6 +58,33 @@
...
+        {
+          "curie": "WB:WBPaper00054725"
+        },
+        {
+          "curie": "WB:WBPaper00056478"
+        },
+        {
+          "curie": "WB:WBPaper00057105"
+        }
+      ]
```
markquintontulloch, line 87: This should be `reference_curies` and then just a list of strings (not objects with curie properties)

test/data/allele_association_ingest_test.json
```diff
@@ -4,7 +4,7 @@
     {
       "curie": "WB:WBVar00000001",
       "allele_symbol_dto": {
-        "name_type_name": "symbol",
+        "name_type_name": "nomenclature_symbol",
```
markquintontulloch, line 7: Need to update `inheritance_mode_name` in this test file as `inheritance_mode_dto` now expected

model/schema/allele.yaml
markquintontulloch: @chris-grove Wouldn't it remain as a `GenomicLocation` object with classes such as `AlleleGenomicLocation` linking it to an allele (or whatever it relates to)

chris-grove: @markquintontulloch 
The current GenomicLocation class is modeled like an association with 'subject', 'predicate', and 'object' slots and only lacks 'evidence', which we probably want anyway, making me think that this should just be an Association subclass where the slots 'subject', 'predicate', and 'object' do not need to be declared.

test/data/allele_test.json
chris-grove: @markquintontulloch I've made the corrections

model/schema/allele.yaml
markquintontulloch: @chris-grove I think I follow you.  So `GenomicLocationAssociation` would have the various properties such as assembly, etc. and AlleleGenomicLocationAssociation would just specify that the subject is an Allele? 

I was thinking that there would be a GenomicLocation object that had a `chromosome` slot rather than an association with Chromosome as the `object`, but you're right - that's not how it's currently modelled so best to stick with that paradigm.  I think there's some redundancy in the modelling here as presumably a Chromosome object would have an assembly, so no need for the `has_assembly` slot on `GenomicLocation`, but I don't think we've really looked at this area yet so there's probably lots of changes/updates in the pipeline.

chris-grove: @markquintontulloch 
Looking at the Neo4J model, assembly is a direct property of the GenomicLocation class and I imagine is based on this agr_schemas model:
https://github.com/alliance-genome/agr_schemas/blob/master/ingest/allele/variantDetail.json
I would be inclined to stick with associating assembly with the genomic location association object unless we are clear on the original logic of why it was structured that way and we can verify we no longer rely on it to be that way.

markquintontulloch: I certainly don't think we should hold up this PR on the basis of sorting out GenomicLocation.

chris-grove:
@markquintontulloch 
OK, I've changed GenomicLocation to GenomicLocationAssociation and *I think* made all the appropriate changes to the model. I still want to add some more test data, but I think the model is in good shape.

markquintontulloch (approved):

PR #184, opened 2022-12-16
Miscellaneous updates for A-Team Dec 2022

chris-grove:
- Add affiliated_alliance_member slot to Person class
- Make 'allele_notes' slot multivalued
- removed 'private_comment' slot
- Make allele_germline_transmission_status (and dto slot) single-valued
- Set domain of 'text_synonyms' to VocabularyTerm

### DIFF
--- a/model/schema/allele.yaml
+++ b/model/schema/allele.yaml
@@ -843,11 +843,13 @@ slots:
     description: Germline transmission status for a given allele
     domain: Allele
     range: AlleleGermlineTransmissionStatusSlotAnnotation
+    multivalued: false
 
   allele_germline_transmission_status_dto:
     domain: AlleleDTO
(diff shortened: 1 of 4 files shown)


model/schema/allele.yaml
```diff
@@ -1056,6 +1058,7 @@ slots:
     description: Notes for a given allele
     domain: Allele
     range: AlleleNoteSlotAnnotation
+    multivalued: true
```
markquintontulloch, line 1061: Should this slot therefore be germline_transmission_statuses?

markquintontulloch: Sorry, missed the hidden lines of code on my phone!

markquintontulloch (approved):

PR #186, opened 2022-12-21
Model reorganization effort attempt 1

chris-grove:
- used allele.yaml as a pilot case
- Moved all DTO classes and DTO-specific slots to allele_dto.yaml
- Alphabetized all class and slot definitions within their respective sections

### DIFF
--- a/model/schema/allele_dto.yaml
+++ b/model/schema/allele_dto.yaml
@@ -1,6 +1,6 @@
-id: https://github.com/alliance-genome/agr_curation_schema/src/schema/allele
-name: Alliance-Schema-Prototype-Allele
-title: Alliance Schema Prototype Allele
+id: https://github.com/alliance-genome/agr_curation_schema/src/schema/allele_dto
+name: Alliance-Schema-Prototype-Allele-DTO
+title: Alliance Schema Prototype Allele DTO
 
 imports:
   - core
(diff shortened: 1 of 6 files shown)


model/schema/allele_dto.yaml
```diff
@@ -0,0 +1,540 @@
...
+  transgene_chromosome_location_curie:
+    description: >-
+      The curie string of the chromosome to which a transgene has been mapped. Used for WormBase
+      transgenes that have been integrated into the genome and mapped to a chromosome.
+    domain: AlleleDTO
+    range: string
+    multivalued: false
+
+## ------------
+## ALLELE DTO SLOT ANNOTATION SLOTS
```
markquintontulloch, line 450: I’m not sure we should separate these from other slots.  It’s not immediately obvious from the slot name that these slots have a range of a SlotAnnotation.  Also, these are slots of the Allele class not the SlotAnnotation classes - my interpretation of ‘Slot Annotation Slots’ would be slots used in the SlotAnnotation classes themselves (eg. mutation_type_curies).  I think just listing all slots alphabetically would be more intuitive.

chris-grove: @markquintontulloch 
You're right, sorry, I thought it made a lot more sense earlier when I was trying to compare the DTO and non-DTO yaml files side-by-side. I've fixed in my latest commit.

chris-grove: Actually, not all done yet...

chris-grove: OK, now fixed

markquintontulloch (approved):
Looks good.  As a side note unrelated to this PR, at some point I think we should go through and give the predicate_name slots.more descriptive names to make it clearer what is expected (as we have for disease_relation_name in the DiseaseAnnotationDTO class)

chris-grove:
@markquintontulloch 
Yep! Sounds good. Thanks!

PR #187, opened 2023-01-10
Add resource descriptors and data provider xrefs

markquintontulloch:


model/schema/resourceDescriptor.yaml
```diff
@@ -0,0 +1,84 @@
+id: https://github.com/alliance-genome/agr_curation_schema/resourceDescriptor.yaml
+name: Alliance-Schema-Prototype-Resource-Descriptor
+title: Alliance Schema Prototype Resource Descriptor
+
+imports:
+  - core
+  - linkml:types
+
```
chris-grove, line 8: Probably need affectedGenomicModel added here as that is where the 'name' slot is defined

markquintontulloch: done

oblodgett (approved):

model/schema/core.yaml
```diff
@@ -893,6 +921,21 @@ slots:
...
+
+  source_organization_abbreviation:
+    domain: DataProviderDTO
+    range: string
+    required: true
+
+  resource_descriptor_page:
+    domain: CrossReference
+    range: ResourceDescriptorPage
+    required: true
```
markquintontulloch, line 937: @oblodgett - I think we need to make this non-required otherwise we would have to create ResourceDescriptor and ResourceDescriptorPage entries for any organisation that the ontologies have xrefs to

chris-grove (approved):

model/schema/core.yaml
oblodgett: Yes I agree.

PR #191, opened 2023-01-19
Molecular mutations zfin

sierra-moxon:
- adding various fields curated specifically at ZFIN 
- unclear why no fields in VariantGenomeLocation are designated as "required" (if this is truly what we want to support for variants, then ZFIN molecular mutations fit in nicely with our variant definitions).


model/schema/variation.yaml
```diff
@@ -101,9 +101,7 @@ classes:
       - reference_sequence
       - variant_sequence
       - consequence
-    slot_usage:
-      evidence_code:
-        required: false
+      - sequence_of_reference_accession_number
```
markquintontulloch, line 104: We have `assembly` and `chromosome` slots on `VariantGenomeLocation` - is this required, and should it really be on the base class?

model/schema/variation.yaml
```diff
@@ -114,6 +112,16 @@ classes:
     slots:
       - assembly
       - chromosome
+      - exon_number
+      - intron_number
+      - variation_strand
+      - number_additional_dna_base_pairs
+      - number_removed_dna_base_pairs
+      - inserted_sequence
+      - deleted_sequence
```
markquintontulloch, line 121: This can be inferred from `reference_sequence` and `variant_sequence` - do we need both pairs of slots?

markquintontulloch (approved):

PR #195, opened 2023-02-14
Gene change event

gildossantos:


gildossantos:
Incorporated feedback from DQM meeting. Names for the SlotAnnotation and slots have been changed. New slots for full_name changes. Removed "withdrawn" event_type option; "deprecation" now covers all cases where object is made obsolete, regardless of whether another object replaces the deprecated object. Tried to expand definitions of event_type terms to clarify how use of this annotation is envisioned in various scenarios.

sierra-moxon (approved):

PR #200, opened 2023-03-13
SCRUM-2719 Add data provider to BiologicalEntity

markquintontulloch:
Add data_provider slot to BiologicalEntity class in order to facilitate loading of non-canonical species by MODs

### DIFF
--- a/model/schema/core.yaml
+++ b/model/schema/core.yaml
@@ -91,6 +91,7 @@ classes:
     slots:
       - curie
       - taxon
+      - data_provider
     slot_usage:
       taxon:
         required: true
@@ -103,6 +104,7 @@ classes:
     slots:
       - curie
       - taxon_curie
+      - data_provider_dto
     slot_usage:
       taxon_curie:
         required: true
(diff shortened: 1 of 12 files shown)


model/schema/core.yaml
```diff
@@ -733,12 +735,14 @@ slots:
...
       sourced and a CrossReference to that organisation's site
     multivalued: false
+    required: true
     range: DataProvider
 
   data_provider_dto:
     description: >-
       Ingest object representing the organization (e.g. MOD) from which the data
       was sourced and a CrossReference to that organisation's site
+    required: true
```
oblodgett, line 745: Does this have to be required? Are we going to set it or are the DQM's going to set it? I think it would be better for us to get it?

markquintontulloch: For disease annotations we require DQMs to set it - I would prefer to keep it like this for everything.  For bulk submission via data loads we can determine the data provider from the bulk load type, but for direct API submissions we can't.  Also, submission by DQMs allows the cross-reference part of data provider to be submitted.

chris-grove (approved):

PR #204, opened 2023-04-11
simplify docker with make commands

sierra-moxon:
from a modified agr_curation_schemas checkout:
`make run-tests` will now:
- build and run a docker container
- install a poetry environment in the docker container
- run the tests using local files.

if the tests fail, you can always
`make remove-container`, edit the necessary files, and run `make run-tests` again

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,9 +27,13 @@ RUN apt-get update \
     && apt-get install --no-install-recommends -y \
     build-essential
 
+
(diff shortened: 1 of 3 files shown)


gildossantos (changes requested):
Test to see if requested change of Dockerfile works for everyone and incorporate that change is it does. If that change (which works for me) breaks things for others, then we can keep as is and just document it in the readme.

PR #205, opened 2023-04-26
A-Team updates: new phenotypes, multiple genetic modifiers, etc.

chris-grove:
- Added several new ontology term classes
- Added 'mod_internal_id' slot definition; slot added to DiseaseAnnotation class
- Added 'synonyms' slot to VocabularyTerm class; updated description of 'synonyms' slot in core.yaml
- Changed 'disease_genetic_modifier' slot to 'disease_genetic_modifiers' and made multivalued

### DIFF
--- a/model/schema/controlledVocabulary.yaml
+++ b/model/schema/controlledVocabulary.yaml
@@ -91,10 +91,15 @@ classes:
       - name
       - vocabulary_description
       - member_terms
+      - synonyms
     slot_usage:
       name:
         required: true
         identifier: true
+      synonyms:
+        description: >-
+          Free text synonym(s) of a controlled vocabulary term
+        required: false
 
   VocabularyTermSet:
     is_a: AuditedObject
--- a/model/schema/controlledVocabulary.yaml
+++ b/model/schema/controlledVocabulary.yaml
@@ -34,13 +34,6 @@ prefixes:
   schema: 'http://schema.org/'
 
 slots:
-  text_synonyms:
-    description: >-
-      Free text synonym(s) of a term, used for controlled vocabulary terms; this is distinct from the
-      'synonyms' slot which has a range of a Synonym class object.
-    domain: VocabularyTerm
-    range: string
-    multivalued: true
 
   member_terms:
     description: >-
@@ -77,20 +70,6 @@ classes:
       - name
       - abbreviation
       - definition
-      - text_synonyms
-    slot_usage:
-      name:
-        required: true
-        identifier: true
-
-  Vocabulary:
-    description: >-
-      A set of VocabularyTerm objects.
-    is_a: AuditedObject
-    slots:
-      - name
-      - vocabulary_description
-      - member_terms
       - synonyms
     slot_usage:
       name:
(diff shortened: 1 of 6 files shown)


model/schema/core.yaml
```diff
@@ -857,6 +858,11 @@ slots:
       The model organism database (MOD) identifier/curie for the object
     range: string
 
+  mod_internal_id:
+    description: >-
+      The model organism database (MOD) internal identifier for the object
+      range: string
```
markquintontulloch, line 864: indentation is off here

model/schema/phenotypeAndDiseaseAnnotation.yaml
```diff
@@ -183,6 +183,7 @@ classes:
       - curie
       - unique_id
       - mod_entity_id
+      - mod_internal_id
```
markquintontulloch, line 186: Need to add this to the DTO class too

model/schema/ontologyTerm.yaml
```diff
@@ -286,9 +285,51 @@ classes:
...
+      PURL: http://purl.obolibrary.org/obo/APO.owl
+
+  DPOTerm:
+    is_a: PhenotypeTerm
+    description: >-
+      A phenotype ontology term from the Drosophila Phenotype Ontology (DPO).
+      PURL: http://purl.obolibrary.org/obo/dpo.owl
+
+  PATOTerm:
+    is_a: OntologyTerm
```
markquintontulloch, line 323: Should this be PhenotypeTerm?  Maybe traits aren't phenotypes but it feels like we probably want to use this ontology in fields that are restricted to phenotype terms.

chris-grove: @markquintontulloch  I am not aware of any curators using PATO terms directly for phenotypes. These terms tend to be used to help generate logical definitions for phenotype terms, but are not used directly as phenotype terms

chris-grove: ...but I can double check with curators

markquintontulloch: cool, we can always change later

markquintontulloch (approved):
Looks good to me

chris-grove:
@markquintontulloch 
Great! Thanks! I'm just going to add an example 'mod_internal_id' to one of the test files and then I'll be done. All tests pass.

model/schema/controlledVocabulary.yaml
```diff
@@ -91,10 +91,15 @@ classes:
...
       - vocabulary_description
       - member_terms
+      - synonyms
     slot_usage:
       name:
         required: true
         identifier: true
+      synonyms:
+        description: >-
+          Free text synonym(s) of a controlled vocabulary term
```
sierra-moxon, line 101: is this supposed to be on "Vocabulary"?  the description talks about vocabulary `term` ?

chris-grove: @sierra-moxon Good catch! Thanks! Fixed

PR #206, opened 2023-05-05
Secondary IDs update

markquintontulloch:
In the current model the `GenomicEntity` class has a `secondary_identifiers` slot, which is multivalued and has a range of `string.`  However, the `Allele` class inherits that slot but also has `allele_secondary_ids` with a range of `AlleleSecondaryIdSlotAnnotation`.  We don't want `Allele` to have both slots, and presumably we want to use the SlotAnnotation paradigm for secondary IDs elsewhere.

In this PR I have removed `secondary_identifiers` from `GenomicEntity` and implemented the SlotAnnotation paradigm for genes and AGMs - we may wish to also do so for other classes that inherit from GenomicEntity or we may wish to stick with the simpler list of strings for those, or not even have secondary IDs at all for some classes.

### DIFF
--- a/model/schema/affectedGenomicModel.yaml
+++ b/model/schema/affectedGenomicModel.yaml
@@ -66,7 +66,7 @@ classes:
       - single_agm
       - secondary_id
     slot_usage:
-      single_allele:
+      single_agm:
         required: true
       secondary_id:
         required: true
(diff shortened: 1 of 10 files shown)


sierra-moxon (approved):

model/schema/affectedGenomicModel.yaml
```diff
@@ -102,6 +119,19 @@ slots:
     range: string
     required: true
 
+  agm_secondary_ids:
+    description: Secondary IDs of a given allele
```
chris-grove, line 123: description: Secondary IDs of a given AGM

markquintontulloch: @chris-grove - Good spot.  Fixed along with another copy-paste error.

PR #210, opened 2023-05-25
Construct updates may 25, 2023

chris-grove:
Several updates to the Construct class and related classes


markquintontulloch (approved):
Looks good to me.  In the alleleDTO.yaml we need to add the properties `inlined: true` and `inlined_as_list: true` to the `construct_component_dtos` slot definition, but that can probably wait as there is more work to do on the DTO side of things.  

We could maybe also switch out `construct_component_reation_enum` for a VocabularyTerm slot now, but that could also wait.

PR #214, opened 2023-06-23
SCRUM-2707 Update Orthology model

markquintontulloch:


oblodgett (commented):
Might want to ask Chris T about this. From what I remember in Neo we created the relationships in the other direction. Subject and object would get switched. Maybe thats not an issue here?

markquintontulloch:
@christabone - do you know anything about Olin's comment above?  From what I can tell from the loader code the genes just go in as gene1 and gene2 with no switching of order.  I don't think it would make any difference since it's a symmetrical relationship, but maybe I'm missing something.

chris-grove:
@markquintontulloch 
I think your PR looks fine so I will approve. I would like to point out that the orthology relationship is not necessarily symmetrical/reciprocal because gene1 may show a strong match to gene2 but the opposite (gene2 to gene1) may not be true (gene2 may clearly match to a different gene much more strongly than gene1). Therefore, there should be two 'GeneToGeneOrthology' objects in the database for each pair of genes with direction being important in each case.

chris-grove (approved):

christabone:
Hi @markquintontulloch -- good question -- it basically works like @chris-grove mentioned. The orthology models we use are asymmetric due to the nature of the algorithms that call the matches. Sometimes an algorithm believes that gene1 -> gene2 is correct but due to some funky biology gene2 -> gene1 isn't always the best match.

PR #215, opened 2023-07-06
Add new ontology terms

markquintontulloch:
@chris-grove I presume all these new terms inherit directly from OntologyTerm


chris-grove (approved):
@markquintontulloch 
Yes, I think direct children of OntologyTerm for these is appropriate.

PR #222, opened 2023-08-30
SCRUM-3156 Phenotype & Disease Annotation remodel and Phenotype DTO classes

markquintontulloch:


oblodgett (approved):
Looks good but i would defer to @chris-grove   for the field lists.

markquintontulloch:
@chris-grove I've stuck with `disease_relation` at the level of the `DiseaseAnnotation` class instead of `relation` in the `Annotation` class for now.  This is how it's currently implemented in the curation app code and a change to the field name would require coordinated changes in the main site code.  Given that phenotype annotations don't have a relation, I think this makes more sense for now anyway.  If we later introduce more Annotation classes and want to have the relation in the parent class then we can refactor at that point.

chris-grove:
@markquintontulloch 
Looks good so far, although I'm not seeing a DTO slot for submitting conditions (like condition_relation_dtos in the DiseaseAnnotationDTO class) for the PhenotypeAnnotationDTO or subclasses. I think we probably need the slot 'condition_relation_dtos' added to PhenotypeAnnotationDTO

markquintontulloch:
> @markquintontulloch Looks good so far, although I'm not seeing a DTO slot for submitting conditions (like condition_relation_dtos in the DiseaseAnnotationDTO class) for the PhenotypeAnnotationDTO or subclasses. I think we probably need the slot 'condition_relation_dtos' added to PhenotypeAnnotationDTO

Good spot.  It should be at the AnnotationDTO level - have fixed it now.

chris-grove (approved):

PR #223, opened 2023-09-04
SCRUM-3230 Construct ingest classes

markquintontulloch:
There are currently a couple of slots that may need to change (`generated_by_identifier` and `manufacture_by_identifier`).  The non-DTO versions of these slots have a range of `Agent`, but there is no common identifying field for `Agent`: the `Person` class has `unique_id`, `Laboratory` has `curie`, `Company` and `Organization` don't have any identifier.  We will need to resolve this in the modelling of the Agent class and subclasses and then update these two field names to be something more descriptive.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -187,6 +187,7 @@ deploy-testpypi:
 
 # datasets used test/validate the schema
 SCHEMA_TEST_EXAMPLES := \
+	construct_test \
 	allele_test \
 	disease_allele_test \
 	disease_agm_test \
--- a/generated/jsonschema/allianceModel.schema.json
+++ b/generated/jsonschema/allianceModel.schema.json
@@ -7278,7 +7278,6 @@
             "required": [
                 "name",
                 "curie",
-                "taxon_curie",
                 "internal"
             ],
             "title": "ConstructDTO",
@@ -16846,7 +16845,6 @@
             },
             "required": [
                 "curie",
-                "taxon_curie",
                 "internal"
             ],
             "title": "ReagentDTO",
--- a/model/schema/affectedGenomicModel.yaml
+++ b/model/schema/affectedGenomicModel.yaml
@@ -47,6 +47,11 @@ classes:
       - sequence_targeting_reagents
       - parental_populations
       - references
+    slot_usage:
+      name:
+        notes: >-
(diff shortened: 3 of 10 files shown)


markquintontulloch:
@sierra-moxon Any idea why this test is failing.  It says ConstructDTO is not defined, but it is and as far as I can tell all imports are in place.  Schema generates fine but then tests fail.

markquintontulloch:
> @sierra-moxon Any idea why this test is failing. It says ConstructDTO is not defined, but it is and as far as I can tell all imports are in place. Schema generates fine but then tests fail.

Fixed this by moving the Construct classes and slots from allele.yaml to reagent.yaml (Construct inherits from Reagent).  But this seems weird - all imports were in place (as far as I can tell) and the schema was generating.  Perhaps the Python generator/compiler can't deal with more than one level of inheritance?  ConstructDTO (which was in allele.yaml) is a ReagentDTO (in reagent.yaml) is an AuditedObjectDTO (in core.yaml).

markquintontulloch:
@chris-grove If curie is not required and Alliance curies will be minted, we probably want to implement unique_Id, mod_entity_id and mod_internal_id fields.

markquintontulloch:
The fact that not all MODs will have curies for constructs (or reagents in general) creates a problem we haven't encountered before.  Until now all classes with associations have hd curies.  As construct associations will be submitted separately from the constructs, we need an identifier to link the association to the construct.  As such we will have to require DQMs to submit either mod_entity_id or mod_internal_id.  The mod_internal_id could simply be a concatenated string, like the AGR-generated unique_id, but it should still be submitted as mod_internal_id to avoid any disparity between the way the concatenated string is generated at the MOD and at the Alliance.  I'm not sure if there's a way to specify this either-or requirement in LinkML in such a way that schema validation would fail if both fields were missing - for now I've added it as a note.

chris-grove (approved):
OK looks good. We'll have to circle back later to address the Agent/generated_by/manufactured_by issue. As for the either/or requirement, I think there are examples Sierra put in here:
https://github.com/alliance-genome/agr_curation_schema/blob/main/examples/rules_examples.yaml
Also, there may be actual examples in our model, but I would have to look further.

PR #225, opened 2023-09-05
Association update proposal

markquintontulloch:
This is what I was thinking along the lines of for Association changes

### DIFF
--- a/model/schema/core.yaml
+++ b/model/schema/core.yaml
@@ -288,14 +288,15 @@ classes:
 
 
   Association:
+    abstract: true
     is_a: AuditedObject
     description: >-
       A typed association between two entities, supported by evidence.  Associations have three base slots: subject,
-      object, and predicate, but they can have any number of additional attributes that help qualify the relationship
+      object, and relation, but they can have any number of additional attributes that help qualify the relationship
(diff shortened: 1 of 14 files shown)


model/schema/core.yaml
```diff
@@ -291,29 +291,42 @@ classes:
...
+    slots:
+      - single_reference
+
+  EvidenceAssociation:
+    abstract: true
+    is_a: Association
+    description: An association supported by any number of InformationContentEntity objects
+    slots:
+      - evidence
+
```
oblodgett, line 329: 👍

oblodgett (approved):

oblodgett:
@markquintontulloch the migration is hurting my head, but I think this is a better model.

markquintontulloch:
> @markquintontulloch the migration is hurting my head, but I think this is a better model.

I agree that it's a better model.  I don't think the migration will be TOO bad because we only have single reference associations at the moment and all annotations will have corresponding entries in that table.  Definitely good to get this sorted before we start trying to implement any of the allele associations though.

sierra-moxon (approved):

chris-grove (approved):
Looks good. I just committed some other replacements of 'predicate' with 'relation' including several controlled vocabulary names

PR #227, opened 2023-09-11
SCRUM-3231 Variant updates and ingest

markquintontulloch:


sierra-moxon (approved):

test/data/variant_association_ingest_test.json
```diff
@@ -2,39 +2,45 @@
   "linkml_version": "1.7.3",
   "variant_genomic_location_association_ingest_set": [
     {
-      "variant_entity_curie": "ZFIN:ZDB-ALT-000412-8",
+      "variant_curie": "ZFIN:ZDB-ALT-000412-8",
       "relation_name": "has_genomic_location",
-      "chromosome_curie": "7",
-      "evidence_code_curie": "ECO:0000006",
```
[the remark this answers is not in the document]
markquintontulloch: There's no evidence_code slot in the non-ingest model.  Do we need to add one?  If so, at what level - VariantGenomicLocationAssociation or GenomicLocationAssociation?

model/schema/variantDTO.yaml
```diff
@@ -36,33 +36,33 @@ emit_prefixes:
...
-      - chromosome_curie
-      - evidence_curies
-      - evidence_code_curie
-      - assembly_curie
-      - start
-      - end
-      - exon_number
-      - intron_number
+      - exon_number # should this be a property of the transcript location? different isoforms could have different values for this
+      - intron_number # should this be a property of the transcript location? different isoforms could have different values for this
```
chris-grove, line 65: I agree that it is odd that these slots (exon_number and intron_number) are added to this class. There are no corresponding slots in the non-DTO model, as far as I can tell, and I agree it likely can't be unambiguous for genomic location unless specifying a particular transcript isoform, if there is more than one.

model/schema/variation.yaml
```diff
@@ -182,16 +193,16 @@ slots:
     exact_mappings:
       - SO:0001889
 
-  dna_mutation_type:
+  dna_mutation_type: # Should the range of this be an object (SOTerm / OntologyTerm) rather than just a curie?
```
chris-grove, line 196: @cmpich @ybradford 
Are the values for this field a ZFIN-specific vocabulary? I see, for example, on these pages:
http://zfin.org/ZDB-ALT-211027-5
Type            Allele with one deletion
http://zfin.org/ZDB-ALT-220726-6
Type            Allele with one insertion
Are these SO Terms or a CV or something else?

model/schema/variation.yaml
```diff
@@ -86,22 +105,20 @@ classes:
       the definition of the transcript or protein isoform used by the source has changed, or if there was an error
       in the source data.
     slots:
+      - hgvs
```
chris-grove, line 108: @markquintontulloch 
Is there a DTO equivalent of the 'hgvs' slot in the VariantLocationAssociationDTO class?

model/schema/variantDTO.yaml
```diff
@@ -36,33 +36,33 @@ emit_prefixes:
...
-      - variant_status
+      - variant_type_curie
+      - note_dtos
+      - source_general_consequence_curie
+      - variant_status_name
+
+  VariantLocationAssociationDTO:
+    abstract: true
+    is_a: GenomicLocationAssociationDTO
+    slots:
```
chris-grove, line 52: I think this needs an 'hgvs' slot

model/schema/variation.yaml
```diff
@@ -118,11 +135,8 @@ classes:
       object:
         required: true
 
-
   VariantTranscriptLocationAssociation:
```
chris-grove, line 138: Now that object has range Transcript, it looks like there's no place to store the Chromosome (it had been the range of object until this level down in the hierarchy). Maybe we need another tag for chromosome of origin? There is a 'chromosome_curie' slot for the VariantLocationAssociationDTO class, but I don't think there is an equivalent chromosome slot in VariantTranscriptLocationAssociation or VariantPolypeptideLocationAssociation

chris-grove (changes requested):
I think we need to resolve the reference to Chromosome from the VariantPolypeptideLocationAssociation and VariantTranscriptLocationAssociation classes. If the object range changes, chromosome now needs a place to go for. Also, I've committed a change to update the slot_usage description of the object slot for VariantPolypeptideLocationAssociation and VariantTranscriptLocationAssociation so that it is clear in the documentation what the range should be (as it currently inherits a note from the parent classes that the object should be a chromosome). Also, we should address this question about exon_number and intron_number for genome level location

model/schema/variation.yaml
markquintontulloch: @chris-grove HGVS IDs have historically been generated rather than submitted

model/schema/variation.yaml
markquintontulloch: @chris-grove I don't think we would need it.  Presumably a transcript would be associated with a gene, which would be located on a chromosome, so this doesn't need to be submitted for the VariantTranscriptLocationAssociation.  I would assume that similarly a protein would be associated with a transcript object....

model/schema/variation.yaml
chris-grove: @markquintontulloch Ah, OK, thanks

chris-grove (approved):

PR #231, opened 2023-09-19
Represent FMS expression submission model in LinkML

gildossantos:


model/schema/expression.yaml
```diff
@@ -69,67 +79,218 @@ classes:
...
       - negated
       - uncertain
-      - associated_with_figure    # this is often redundant
-      - related_notes
-
+      - associated_with_figures
+      - expression_annotation_image_associations
     slot_usage:
-      associated_with_figure:
+      subject:
```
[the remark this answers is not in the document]
gildossantos: These have been changed to `expression_annotation_subject` and `expression_pattern`.

model/schema/expression.yaml
```diff
@@ -69,67 +79,218 @@ classes:
...
-      - where_expressed
-      - expression_qualifiers
+      # from_expression_experiment   # For DTO only; for database, relate via non-DTO ExpressionExperiment.expression_annotations instead.
+      - where_expressed_statement
       - negated
       - uncertain
-      - associated_with_figure    # this is often redundant
-      - related_notes
-
+      - associated_with_figures
```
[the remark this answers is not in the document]
gildossantos: Changed to `related_figure(s)`.

model/schema/expression.yaml
```diff
@@ -233,58 +421,60 @@ slots:
     domain: TemporalContext
     range: StageTerm
     multivalued: false
+    required: true
```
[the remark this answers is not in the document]
gildossantos: I've had a closer look at the FMS requirements. I think I've updated to be mostly in line now - see `notes` for the `GeneExpressionAnnotation`. I should note, however, that the `AnatomicalSite` postconditions are still not quite as restrictive as the FMS `whereExpressed` constraints - see `* NOTE: For the FMS whereExpressed component ...` for details.

model/schema/expression.yaml
```diff
@@ -44,104 +44,258 @@ classes:
...
+      image information).
     slots:
-      - belongs_to_expression_experiment # remove and model as a collection on the class.
-      - when_expressed
-      - where_expressed
-      - expression_qualifiers
+      # from_expression_experiment   # For DTO only; for database, relate via non-DTO ExpressionExperiment.expression_annotations instead.
+      - expression_annotation_subject
+      - expression_pattern
+      - expression_assay_used
```
gildossantos, line 117: @markquintontulloch @abecerra @chris-grove
*** NOTE - new change**. I have just added `expression_assay_used` to the `ExpressionAnnotation` object. This slot is redundantly present in the `ExpressionExperiment` object. My original rationale for keeping the slot only in the expt object was that I imagined DQMs would be submitting both types of objects. Now that the A-Team is simply mapping FMS data to LinkML, the old `assay_used` slot was the only one not present in the primary `ExpressionAnnotation` object, and it seemed unnecessarily complicated to have A-Team derive `ExpressionExperiment` objects just so they had a place to put the FMS `assay` data; any `ExpressionExperiment` objects that the A-Team could derive would also be very scant compared to what a DQM could provide (e.g., IDs for these expts). Anyway, just a suggestion but we can change that back if preferred.

model/schema/expression.yaml
```diff
@@ -69,67 +79,218 @@ classes:
...
+        range: Gene
+      # ACTION ITEM: Create is_expressed_in VocabularyTerm (Daniela).
+      relation:
+        description: >-
+          For a gene, the relation should be constrained to the expresssed_in
+          VocabularyTerm.
+
+  # DQM NOTE: Each FMS expression annotation is limited to one stage. However,
+  #           in the LinkML model, it's possible to specify a range.
+  ExpressionPattern:
```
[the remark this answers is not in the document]
markquintontulloch: Why with a uniqueId? A single expression annotation has a single ExpressionPattern attached, and that has a single TemporalContext and single AnatomicalSite.  You're either creating the objects or updating the ones attached to the annotation - there's no lookup to be done.

model/schema/expression.yaml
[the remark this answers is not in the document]
markquintontulloch: Less restrictive is not a problem, for loading from the FMS it's only an issue if the LinkML model requires a field that the FMS doesn't require

model/schema/expression.yaml
markquintontulloch: @abecerra I presume the intention is to create ExpressionExperiment objects from the FMS submission, so this duplication is not required?

model/schema/expression.yaml
```diff
@@ -44,104 +44,255 @@ classes:
...
+      The experimental design used to determine the expression of some entity,
+      defined by the entity_assayed, the specimen, the assay, the reagents
+      (e.g., Antibody, Probe), the environmental conditions, and the reference.
+      It groups ExpressionAnnotations; it may have its own notes.
     notes: >-
-      Specimens are defined by their AGM and/or alleles (or gel lane at MGI).
+      Specimens are defined by their AGM (or gel lane at MGI).
+      This entity is originally specific to WB and MGI.
     slots:
+      - curie
```
markquintontulloch, line 57: curie, mod_entity_id, mod_internal_id, and data_provider are all inherited from SubmittedObject so don't need to be defined here

model/schema/expression.yaml
```diff
@@ -44,104 +44,251 @@ classes:
...
+      - expression_annotations    # For non-DTO only; for data submission, relate via ExpressionAnnotationDTO.from_expression_experiment instead.
     rules:
       - postconditions:
           any_of:
             - slot_conditions:
-                biological_entity_assayed:
+                entity_assayed:
                   required: true
             - slot_conditions:
                 reagents_used:
```
chris-grove, line 74: reagents_used needs to be swapped with detection_reagents

model/schema/expression.yaml
chris-grove: @gildossantos OK makes sense to me. I'll let @abecerra address Mark's question. I think the question comes down to whether or not we have sufficient information in the FMS EXPRESSION data to infer all defining properties of an ExpressionExperiment, which I can't answer off hand. Ideally, we would create the ExpressionExperiment objects now so we don't have to create them later and then transfer the data from one table to the other.

model/schema/expression.yaml
```diff
@@ -44,104 +44,251 @@ classes:
...
+  # DQM NOTE: Each FMS expression annotation is limited to one stage. However,
+  #           in the LinkML model, it's possible to specify a range.
+  ExpressionPattern:
+    is_a: AuditedObject
+    description: >-
+      An expression pattern, as defined by spatial and temporal attributes,
+      including all qualifiers.
+    slots:
+      - when_expressed
+      - where_expressed
```
chris-grove, line 244: I see that where_expressed is considered a required slot but when_expressed is not. Is it true that an AnatomicalSite would always be required, but a TemporalContext would not? Maybe this is meant to have a post_conditions statement indicating that either one or the other is required?

model/schema/expression.yaml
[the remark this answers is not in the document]
abecerra: I meant, the uniqueId would be for the annotation. The ExpressionPattern, TemporalContext and AnatomicalSite would be loaded and stored

model/schema/expression.yaml
abecerra: Yes, the duplication is not required, Thanks @gildossantos! We would create the ExpressionExperiments when loading @markquintontulloch, thanks!

chris-grove (approved):

abecerra (approved):

PR #234, opened 2023-09-28
Update BulkLoadFile class

markquintontulloch:


oblodgett:
What are these for? Maybe I missed a discussion somewhere?

markquintontulloch:
@oblodgett linkml_schema_version is already in our Java model.  There was a request to track which MOD release corresponds to each submitted file.  This slot has already been added to the Ingest class, but it needs to be stored on the bulkLoadFile too.

chris-grove (approved):

PR #237, opened 2023-10-19
align with FMS submission: add gene_synopsis, make gene_type required

gildossantos:
Simple changes to bring LinkML model in line with agr_schemas model.
1. Add gene_synopsis slot (a simple string).
2. Make gene_type/gene_type_curie (aka soTermId, gene biotype) a required slot for Gene/GeneDTO classes

I want to handle gene genomic locations in a separate PR.

### DIFF
--- a/generated/jsonschema/allianceModel.schema.json
+++ b/generated/jsonschema/allianceModel.schema.json
@@ -10837,6 +10837,12 @@
                     "description": "Classifies the entity as private (for internal use) or not (for public use).",
                     "type": "boolean"
                 },
+                "note_dtos": {
+                    "items": {
+                        "$ref": "#/$defs/NoteDTO"
+                    },
+                    "type": "array"
+                },
                 "obsolete": {
                     "description": "Entity is no longer current.",
                     "type": "boolean"
--- a/generated/jsonschema/allianceModel.schema.json
+++ b/generated/jsonschema/allianceModel.schema.json
@@ -10410,6 +10410,10 @@
                     },
                     "type": "array"
                 },
+                "gene_synopsis": {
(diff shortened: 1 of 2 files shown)


chris-grove:
@gildossantos This looks OK, but I'd like to point out that the mechanism for gene descriptions we worked out with the gene descriptions group was to submit various description types as related_notes with an appropriate note_type for each type of description. If you check the "Gene Related Note Type" vocabulary in the curation tool:
https://curation.alliancegenome.org/#/vocabterms
We already have note types "automated_gene_description" and "MOD_provided_gene_description" for the gene description group use cases, in addition to the gene note types requested by SGD:
`
gene_regulation_summary
gene_disease_summary
gene_protein_summary
gene_interaction_summary
gene_function_summary
gene_phenotype_summary
gene_sequence_summary
`
We probably don't want to have multiple different ways of attaching a gene description, so we may want to either forgo the "gene_synopsis" slot OR discuss a different approach with the gene descriptions and gene working groups.

chris-grove (changes requested):
I think we should not include a gene_synopsis slot in favor of using related_notes with different note types depending on the type of description/synopsis

gildossantos:
@chris-grove That makes sense. I removed the `gene_synopsis` slots. In the Gene notes, I added some explanation on the mapping of FMS BGI `geneSynopsis` and `geneSynopsisUrl` to LinkML slots that I think will be helpful to DQMs, at least during the FMS-LinkML transition. I added `note_dtos` to the `GeneDTO` slot to allow for notes submission.

markquintontulloch (approved):

chris-grove (approved):
OK, looks good, thanks @gildossantos !

PR #238, opened 2023-10-23
more abstract location object; reformat variant location objects; gene genomic locations

gildossantos:


model/schema/core.yaml
```diff
@@ -517,16 +521,34 @@ slots:
...
   end:
-    description: The end of the feature in positive 1-based integer coordinates
+    description: >-
+      The end of the feature in positive 1-based integer coordinates relative
+      to the reference landmark.
+    domain: LocationAssociation
     range: integer
 
+  # QUESTION: How to specify strandedness for genomic location.
+  strand:
```
sierra-moxon, line 538: https://github.com/biolink/biolink-model/blob/db44be0c49939229c28cbb71a715127941e0ce0b/biolink-model.yaml#L5711C1-L5711C1

https://github.com/biolink/biolink-model/blob/db44be0c49939229c28cbb71a715127941e0ce0b/biolink-model.yaml#L11264 

?

gildossantos: Perfect. Thank you!

model/schema/core.yaml
```diff
@@ -355,71 +356,74 @@ classes:
...
+      together related chromosome sequences from different genome assemblies.
+    notes: >-
+      I'm trying to distinguish the idealized biological entity from a species'
+      genome from a chromosome sequence assembly that is specific to a given
+      genome assembly version.
+    exact_mappings:
+      - SO:0000340
+      - GO:0005694
+
+  ChromosomeAssembly:
```
markquintontulloch, line 374: I wonder if the name should be something more generic, e.g. AssemblyComponent.  I'm thinking that in the various places that we are likely to use this we are probably going to really mean scaffolds or contigs in addition to chromosomes.

model/schema/core.yaml
```diff
@@ -355,71 +356,74 @@ classes:
       - prefix_order
 
   Chromosome:
-    is_a: AuditedObject
-    description: The ID of the landmark used to establish the coordinate system for the current feature.
+    is_a: GenomicEntity
```
markquintontulloch, line 359: Should this really be a GenomicEntity?  If we want something to represent the idealized biological entity across different assemblies then will it have a curie, for example?

model/schema/core.yaml
```diff
@@ -355,71 +356,74 @@ classes:
...
+      all of a chromosome's DNA sequence. In the context of this model, this
+      entity is used as a landmark to establish a coordinate system for the
+      location of other genomic entities.
     slots:
-      - curie
       - display_name
-    slot_usage:
-      curie:
-        required: true
+      - has_assembly
```
markquintontulloch, line 383: I'd prefer this to just be called `assembly` - `has_assembly` sounds like the name of a boolean field to me

markquintontulloch (commented):
Looks great on the whole - just a couple of picky things and thoughts.

model/schema/core.yaml
gildossantos: I agree and I'll make the change. Going with this new definition: A` scaffold component of a genome assembly, including chromosomal pseudo-molecule sequences, unplaced/unlocalized scaffolds and contigs, and organelle genomes.` On NCBI Assembly pages, the generic term `scaffold` includes the chromosomal assemblies, so I used that in the definition, but I can change that if that seems wrong.

model/schema/core.yaml
gildossantos: Using the 1920s definition, I would say yes, but from a modern practical definition, perhaps it's more confusing so I'll just change it to BiologicalEntity for now.

model/schema/core.yaml
gildossantos: I'll change this to `genome_assembly`.

markquintontulloch: yeah, I forgot slot names can't be the same as class names - `genome_assembly` sounds good

model/schema/core.yaml
markquintontulloch: BiologicalEntity objects still require a curie field though - will these have a curie?

markquintontulloch: Sounds good, and some sort of name/label slot

gildossantos:
@markquintontulloch @chris-grove @sierra-moxon I've tried to address the suggestion from today's DQM meeting (creating sibling curated and source variant location objects with different requirements), as well as issues raised in last week's variants working group meeting (making `source_general_consequence` a `SlotAnnotation`). Please have a look. Thanks!

chris-grove:
@gildossantos  Yeah, more or less. Similar to our parallel constraint differences for "Curated" vs. "Source" location associations, it could satisfy the curators requirements if curators were able to (1) indicate a generic chromosome/scaffold/etc. object without specifying a specific chromosome or assembly or (2) provide the same with but with a specific chromosome and/or assembly, depending on the case.

gildossantos:
As described above, would we use these `Chromosome` (child of `AssemblyComponent`) entities as objects in cytological or genetic location associations? That would be a simple solution, but that does not seem quite right to me conceptually. Is it just me?

gildossantos:
@markquintontulloch - yes. Would it work to loosen the `genome_assembly` requirement, or, allow for a placeholder GenomeAssembly entity in that slot?

markquintontulloch:
If we were to use a placeholder then I think that brings us back to the idea of 'unspecified' genomic objects and we wouldn't need to make Chromosome a child of GenomeAssembly.  I'm in favour of that as a simple(ish) solution, but what it does mean is that we have to create:
a) an unspecified assembly (possibly one for each MOD/species)
b) specific chromosome objects for each chromosome that belongs to that assembly so that you can refer to e.g. chromosome 2 of unspecified assembly
c) unspecified chromosome objects for each assembly as it seems to be a requirement that curators might want to place a genomic entity on a specific assembly but an unspecified chromosome

I'm not sure that poses too many problems, although it does raise the issue of how we would prevent those unspecified objects being used for curated locations yet allowed in source locations - maybe that just has to be some business logic in the code and not specified by the model.

Loosening the requirement for genome_assembly would essentially have the same result.  We would need chromosome objects that aren't linked to an assembly in order to be able to link genomic entities to them in the database.  I think I'd prefer the assembly to be required and go down the route of unspecified objects though as it reduces the likelihood of things being assigned to an unspecified assembly when the assembly is actually known.

If we do opt for unspecified assembly/chromosome objects then I don't know how you feel about keeping the Chromosome class as currently implemented.  I like the idea of having an association between AssemblyComponents and Chromosomes so that you could pull out all the features that are positioned on e.g. Chromosome 3 of any mouse assembly, including the unspecified mouse assembly.  Whether or not there is a valid use case for that though is probably a question for curators.

All of this seems like a lot of hoops to jump through in order to be able to store source locations where there isn't even enough information to place something on a chromosome at the point of curation.  I'd argue that we should just not create genomic locations for such things - if the reference is stored then the curator can just refer to that if they want to try and assign a curated location at a later date.  That doesn't seem to be a route the curators want to go down though.

gildossantos:
@markquintontulloch - all good points. I did like the idea of associating assembly components across genome assemblies, but I don't think we have a use case for that. That's why I asked about what we would use as the object in genetic and cytological mapping data, because that was the main motivation in my original conception of `Chromosome`. If we want to use `AssemblyComponent` or some child of that as the object for genetic and cytological mapping associations (which I think is easiest but a slight mis-fit), then I think we have a way forward). Otherwise, I think we need to hold onto the `Chromosome` as I defined it (not make `Chromosome` a child of `AssemblyComponent`), but we make new `AssemblyComponent` objects having unspecified assembly for each of the chromosomes for each species, to be used for less specific genomic mapping requirements (as you've laid out).

markquintontulloch:
@gildossantos I agree that using AssemblyComponent for genetic and cytological mapping data doesn't really fit right.  I like the idea of keeping the current Chromosome objects for the purpose of having mapping location classes where the object is Chromosome.

model/schema/allele.yaml
```diff
@@ -1078,7 +1078,7 @@ slots:
       The chromosome to which a transgene has been mapped. Used for WormBase transgenes
       that have been integrated into the genome and mapped to a chromosome.
     domain: Allele
-    range: Chromosome
+    range: AssemblyComponent
```
chris-grove, line 1081: WB does not have specific assembly information for these locations. They are just roman numerals indicating the chromosome identity. If WB is the only one that will use this property, this would probably be better left as range: Chromosome. If others will use it and want to be able to indicate an assembly then we can discuss an option that doesn't require an assembly.

chris-grove (approved):
I think it looks good @gildossantos 
I would request changing the range of transgene_chromosome_location back to Chromosome, so curators have the option of being generic (not having to specify an assembly), but we can sort that out later and/or assert the latest assembly by default.

model/schema/allele.yaml
gildossantos: @chris-grove That makes sense, and FlyBase has some data that falls into that category as well. I've reverted the change.

chris-grove:
@gildossantos Is this ready to become a non-draft PR? We can merge it then

gildossantos:
@chris-grove - Yes, I think so.

PR #240, opened 2023-11-13
Update variantDTO.yaml

chris-grove:
Changed description of VariantDTO class to refer to Variant class instead of Allele class


sbello (approved):
Thanks for the change Chris

PR #242, opened 2023-11-21
Curie refactor

markquintontulloch:
As discussed at the DQM/Modelling meeting on 21st Nov

### DIFF
--- a/model/schema/core.yaml
+++ b/model/schema/core.yaml
@@ -97,32 +97,12 @@ classes:
     slots:
       - mod_entity_id
       - mod_internal_id
-    rules:
-      - postconditions:
-          any_of:
-            - slot_conditions:
-                mod_entity_id:
-                  required: true
-            - slot_conditions:
-                mod_internal_id:
-                  required: true
-          description: At least one of mod_entity_id or mod_internal_id should be populated.
 
   SubmittedCurieObjectDTO:
     is_a: AuditedObjectDTO
     slots:
       - mod_entity_id
       - mod_internal_id
-    rules:
-      - postconditions:
-          any_of:
-            - slot_conditions:
-                mod_entity_id:
-                  required: true
-            - slot_conditions:
-                mod_internal_id:
-                  required: true
-          description: At least one of mod_entity_id or mod_internal_id should be populated.
 
   BiologicalEntity:
     is_a: SubmittedCurieObject
--- a/model/schema/core.yaml
+++ b/model/schema/core.yaml
@@ -101,11 +101,11 @@ classes:
       - postconditions:
           any_of:
             - slot_conditions:
-              mod_entity_id:
-                required: true
+                mod_entity_id:
+                  required: true
             - slot_conditions:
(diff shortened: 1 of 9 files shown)


oblodgett (commented):
I would have guessed this refactor would have been more than this?

markquintontulloch:
> I would have guessed this refactor would have been more than this?

I don't think so, although I may well have missed something.  Most things with a curie, mod_entity_id, or mod_internal_id inherit from something else, so it's only a few parent classes that are changing.  The DTO classes are largely unaffected.

The Java side on the other hand..... well, the migration will be fun!

model/schema/core.yaml
```diff
@@ -82,28 +82,65 @@ classes:
...
+  SubmittedCurieObject:
+    is_a: CurieAuditedObject
+    description: >-
+      Parent class for all LinkML classes of submitted entities or annotations
+      that have (or will have) a curie generated at the Alliance
+    slots:
+      - mod_entity_id
+      - mod_internal_id
+    rules:
+      - postconditions:
```
chris-grove, line 101: @markquintontulloch 
We need to consider the fact that currently the DiseaseAnnotation class has fields for curie, mod_entity_id, and mod_internal_id but SGD, ZFIN, RGD, and OMIM (via RGD) do not provide any of those fields. We only have a generated internal UniqueID. This was a requirement of disease annotation submission discussed early on.

markquintontulloch: Sure, will remove the rules to cover that case

chris-grove:
@markquintontulloch @oblodgett 
I think the challenging bit here is that we still have (and can likely expect to have) additional classes that have curies but inherit from classes that do not (and arguably should not always) have a curie. For example, the following classes still have curie slots:
```
Laboratory
GeneNomenclatureSet
Identifier
GeneInteraction
GeneInteractionDTO
Annotation
```
It would take considerable reconfiguration of the model to make it so that these classes all inherit from a chain of ancestors that all have a curie. We could opt to add a curie field in some classes that do not inherit from CurieAuditedObject, but then we would modify the definition of CurieAuditedObject so that it is not declared to be the parent/root class of ALL classes with a curie. What do you think?

markquintontulloch:
@chris-grove I may be wrong (@oblodgett can correct me), but I don't think the issue with Hibernate would prevent us from adding the curie field to a child class when the parent doesn't have a curie.  As all of the entities would have the same base class (AuditedObject), it wouldn't matter where in the hierarchy curie was implemented.  The issue now is that in Java the base class of Construct is Reagent (it inherits from AuditedObject but this isn't implemented as a table), while the base class of genomic entities is BiologicalEntity.

chris-grove (approved):

markquintontulloch:
@chris-grove Also moved `data_provider` from `Reagent` and `BiologicalEntity` into `SubmittedObject` - I think we will want this for all submitted classes.

sierra-moxon (approved):

PR #243, opened 2023-12-06
Updates to submit Transcripts, Chromosomes, Proteins, etc.

chris-grove:


gildossantos (approved):
Thanks @chris-grove. This looks great. I just have some minor questions about where NCBI RefSeq/GenBank/Assembly curies fit into the DTO models for transcripts, proteins, assembly components and genome assemblies.

model/schema/core.yaml
```diff
@@ -417,6 +436,15 @@ classes:
       - genome_assembly
       - maps_to_chromosome
 
+  AssemblyComponentDTO:
+    is_a: GenomicEntityDTO
+    description: >-
+      Ingest class for assembly components
+    slots:
+      - display_name
+      - genome_assembly_curie
```
markquintontulloch, line 445: The new paradigm for DTO slots referring to classes inheriting from SubmittedObject is to use _identifier as the suffix rather than _curie

model/schema/core.yaml
```diff
@@ -426,11 +454,20 @@ classes:
...
       specimen_genomic_model:
         description: >-
           The strain or genotype of origin for the genome sequence.
 
+  GenomeAssemblyDTO:
+    is_a: BiologicalEntityDTO
+    description: >-
+      Ingest class for genome assemblies
+    slots:
+      - specimen_genomic_model_curie
```
markquintontulloch, line 468: The new paradigm for DTO slots referring to classes inheriting from SubmittedObject is to use _identifier as the suffix rather than _curie

markquintontulloch (changes requested):

model/schema/core.yaml
```diff
@@ -663,6 +714,14 @@ slots:
     required: true
     multivalued: false
 
+  genome_assembly_curie:
```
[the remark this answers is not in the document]
chris-grove: Yes, sure. We should agree upon the appropriate format standards for any of these curie or cross reference submissions. For now, I was just throwing in fake test data. We could also add your point as a note.

model/schema/core.yaml
```diff
@@ -417,6 +436,15 @@ classes:
       - genome_assembly
       - maps_to_chromosome
 
+  AssemblyComponentDTO:
```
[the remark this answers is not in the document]
chris-grove: @gildossantos Thanks for taking a look! I had a similar question this morning for @markquintontulloch regarding where a DQM should submit, for example, a UniProt ID as the primary ID for a protein when the MOD does not have its own curie/ID. We think submitting to the "mod_entity_id" field is fine for now, with the interpretation being that is the identifier uses for the object currently rather than necessarily a MOD's own identifier. If/when we mint Alliance AGRKB curies for this and other classes, we can move the, for example, UniProt ID into the cross references field and leave the mod_* fields blank. I agree we want to be careful not to make it look as if we're claiming that a MOD is claiming an external ID as their own

model/schema/core.yaml
chris-grove: Fixed

model/schema/core.yaml
chris-grove: Fixed

model/schema/core.yaml
markquintontulloch: We should probably update the description in the slot definition too so that it is clear that a MOD/external curie or internal ID can be used

chris-grove: Yep, good point. I've updated the description of mod_entity_id now

markquintontulloch: Sorry, I should have been clearer. I meant the descriptions of the two _identifier slots in this PR - they currently say 'curie', which may be misleading.

PR #245, opened 2024-02-12
Base-table-specific subject and object slots

markquintontulloch:
@chris-grove @oblodgett
One of the reasons for the LinkML v2.0.0 migration was because Hibernate couldn't deal with Association subject/object slots with ranges belonging to different base tables in the database (e.g. one subject being a Gene with base table BiologicalEntity, and another being Construct with a base table Reagent).

The initially proposed solution was to have a single base table (AuditedObject) so that anything could be used for any slot.  When I implemented this we ran into the problem (that we can't currently get around) of the Hibernate indexing queries resulting in a number of columns that exceeds the Postgres limit.

The current PR in the agr_curation repo avoids this issue by making CurieObject the base table.  This allows OntologyTerms, Reagents, and BiologicalEntities to be used in the subject/object fields - this covers all our currently implemented associations but not all of the associations in the LinkML model.  Also, having implemented this in Java, it makes database queries, loading and indexing significantly slower.  The speed issue might be something we can improve, but that still leaves us with associations in the LinkML model that cannot be implemented as the model stands.

Given all of the above, I think we need to have separate subject and object slots for each base table in the database (this PR).  This is how we are currently working around the issue in the curation system, with a departure from the LinkML model version that we are reporting that we support.  I don't love the duplication, but I think that is something we have to live with.  I also don't like the fact that you need to know the base table in the database to determine the appropriate slot (e.g. when a Construct is the subject of an association, you need to know that the corresponding base table in our DB is the reagent table and therefore the slot required is subject_reagent).  For that reason, I'm inclined to think that maybe we should go even further with the separation/duplication than in this PR and have class-specific subject and object slots (e.g. subject_gene, subject_allele, etc.), although that would make life a bit harder when we want different things in the same column (e.g. for disease annotation subject).

Any thoughts?

### DIFF
--- a/generated/jsonschema/allianceModel.schema.json
+++ b/generated/jsonschema/allianceModel.schema.json
@@ -16895,6 +16895,10 @@
                     "description": "Classifies the entity as private (for internal use) or not (for public use).",
                     "type": "boolean"
                 },
+                "object_ontology_term": {
+                    "description": "connects an association to the object of the association when the object is a child of OntologyTerm. For example, in a gene disease annotation, a DOTerm is the object describing the disease.",
+                    "type": "string"
+                },
                 "obsolete": {
                     "description": "Entity is no longer current.",
                     "type": "boolean"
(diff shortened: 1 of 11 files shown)


sierra-moxon:
If we go this way, we might consider using the "attribute" syntax in LinkML for these slots (it's just a syntax change, doesn't solve any problems, but it keeps the slots in the class they refer to and reduces the size of the model file itself.  And it makes it clear that the slots shouldn't be reused for other classes).

markquintontulloch:
@sierra-moxon Maybe I don’t understand the use of attributes, but we’d still want to allow the slots to be used in different classes

PR #246, opened 2024-02-15
Refactor association slots using domain prefix

markquintontulloch:
Alternative version of association slot refactor using domain class prefix to differentiate, as discussed in sprint planning.  As per Sierra's suggestion, I've used "attributes" where possible to define slots inline in the classes if reuse of the slot is not intended.

### DIFF
--- a/model/schema/geneInteraction.yaml
+++ b/model/schema/geneInteraction.yaml
@@ -64,6 +64,8 @@ classes:
           the object gene in the association. If the relation is symmetric,
           subject vs object is arbitrary. We allow a gene product to stand
           as a proxy for the gene or vice versa.
+    slots:
+      - gene_association_subject
     slot_usage:
       gene_association_subject:
         description: >-
@@ -73,7 +75,7 @@ classes:
 #    exact_mappings:
 #      - biolink:GeneToGeneAssociation
 
-  GeneToGeneAssociationDTO:
+  GeneGeneAssociationDTO:
     is_a: AuditedObjectDTO
     slots:
       - subject_gene_identifier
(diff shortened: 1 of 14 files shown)


gildossantos (approved):

oblodgett (approved):

sierra-moxon (approved):

model/schema/geneInteraction.yaml
```diff
@@ -50,33 +50,30 @@ classes:
...
+        required: true
         description: >-
-          the subject gene in the association. If the relation is symmetric,
+          the object gene in the association. If the relation is symmetric,
           subject vs object is arbitrary. We allow a gene product to stand
           as a proxy for the gene or vice versa.
-      object:
-        range: Gene
+    slot_usage:
+      gene_association_subject:
```
chris-grove, line 68: I see that "gene_association_subject" is declared in GeneGeneAssociation slot_usage and as a defining slot in GeneInteraction, but the slot itself is never actually instantiated in any class slot list. I suppose it would have gone in a more generic GeneAssociation class above GeneGeneAssociation, but that class does not exist. Do we need a generic GeneAssociation class? In any case, the slot "gene_association_subject" needs to be declared as a slot in some base class. @sierra-moxon Do we have any LinkML pipeline checks to make sure that if a slot is considered defining for a class, then the slot must have been instantiated as a slot for that class or some superclass?

markquintontulloch: Good spot, slot definition now added.  A GeneAssociation class with the gene_association_subject slot as an attribute isn't really an option as we have different gene associations that inherit from LocationAssociation, EvidenceAssociation, and SingleReferenceAssociation.

PR #251, opened 2024-04-06
Remove comment on phenotype annotation SGD strain background

chris-grove:


chris-grove:
Needs further discussion about where to place an sgd_strain_background (or just strain_background) slot for phenotype annotations: SGD needs at both gene and allele phenotype level, WB at the allele level (and for RNAi annotations). Could add to PhenotypeAnnotation super class but need to consider impact. Also, consider changing slot name to "strain_background" with a note in the model to indicate this is reserved for SGD background strain. Then this can be repurposed for strain background in other MOD phenotype annotation (like WB, at least).

PR #253, opened 2024-05-17
Renamed orthology to homology and added paralogy classes and fields

VarunReddy1111:


model/schema/homology.yaml
```diff
@@ -59,79 +59,104 @@ classes:
...
+  
+  similarity:
+    description: Similarity score between the two genes
+    range: integer
+    multivalued: false
+    
+  identity:
+    description: Identity score between the two genes
+    range: integer
+    multivalued: false
```
oblodgett, line 161: multivalued defaults to false so this can be left off of these?

PR #254, opened 2024-06-19
SCRUM-4173 Update model for GFF load

markquintontulloch:


oblodgett (approved):

markquintontulloch:
> So on the java side I would still implement this as a `private SOTerm variantType;` otherwise we will need to create a join table, etc.

Yeah, that's the intention.  The `is_a: genomic_entity_type` is just to avoid defining almost identical slots for the various genomic entities

chris-grove (approved):

PR #255, opened 2024-06-20
SCRUM-4090 Added required fields in homology

VarunReddy1111:


sierra-moxon:
@VarunReddy1111, @chris-grove  - is it possible to also link this change to an issue in Jira? -- does that happen automatically from the other end (E.g. I would see the link in the Jira issue but maybe not here)?

chris-grove:
OK, I've labeled both PRs with the ticket ID, so they are both linked now.

chris-grove (approved):
Artifacts are generated and tests pass without issue, approved

PR #258, opened 2024-07-01
HTP expression data modeling first pass

chris-grove:
Here is a first pass at building the LinkML model for the HTP expression data currently submitted to the FMS and stored in Neo4J. Basically a LinkML version of everything found here:

https://github.com/alliance-genome/agr_schemas/tree/master/ingest/htp

The pipeline generates the JSON schema artifact without error and all existing tests pass without error.

There is a stub test file for high throughput expression, high_throughput_expression_test.json that can be fleshed out later.


model/schema/highThroughputExpression.yaml
```diff
@@ -0,0 +1,578 @@
...
+
+
+classes:
+
+  HighThroughputExpressionDatasetAnnotation:
+    is_a: SubmittedObject
+    description: >-
+      An entry for a published high-throughput expression dataset annotation.
+    slots:
+      - dataset_id # datasetId in agr_schemas json
```
markquintontulloch, line 46: Do we need this?  SubmittedObject already has `mod_entity_id` and `mod_internal_id`.

model/schema/highThroughputExpression.yaml
```diff
@@ -0,0 +1,578 @@
...
+      - name # title in agr_schemas json
+      - references # publications in agr_schemas json
+      - related_note # use 'summary' note type; summary in agr_schemas json
+      - number_of_channels # numChannels in agr_schemas json
+      - sub_series # subSeries in agr_schemas json
+      - category_tags # categoryTags in agr_schemas json
+    slot_usage:
```
markquintontulloch, line 53: dataset_id, name, and category_tags are all required fields in the agr_schema repo - should they be here too?
dateAssigned is also a required field in the old repo but I don't see an equivalent field here, is that intentional?

model/schema/highThroughputExpression.yaml
```diff
@@ -0,0 +1,578 @@
...
+          primaryId and secondaryId
+      genetic_sex:
+        description: >-
+          The sex of the organism the sample came from
+      expression_assay_used:
+        description: >-
+          ID of the type of assay used for this sample; sample or experiment property? See enumeration below.
+        range: high_throughput_expression_assay_values
+    rules:
+      - postconditions:
```
markquintontulloch, line 121: I think it would be clearer to just make `datasets_ids`, `expression_assay_used`, and `sample_type` all required fields and then just have this post-condition to indicate that either `mod_entity_id` or `sample_title` is required

markquintontulloch: Would `mod_internal_id` also be valid in the absence of `mod_entity_id` or `sample_title`?

model/schema/highThroughputExpression.yaml
```diff
@@ -0,0 +1,578 @@
...
+    range: integer # enum [1,2]
+    multivalued: false
+
+  dataset_id: # primaryId in agr_schemas htpId.json
+    description: >-
+      The ID from the data provider for the dataset. When available,
+      this is the GEO ID. When no GEO ID is available, it will be the
+      ArrayExpress ID. When these both are unavailable, it will be the
+      MOD ID.
+    range: uriorcurie
```
markquintontulloch, line 409: In the old schema this is an object (schema in htpId.json).  Should this really have a range of `HighThroughputExpressionDatasetID`?

model/schema/highThroughputExpression.yaml
```diff
@@ -0,0 +1,578 @@
...
+                  required: true
+                expression_assay_used:
+                  required: true
+                sample_type_curie:
+                  required: true
+          description: >-
+            At least sample_id, dataset_ids, expression_assay_used, and sample_type should be populated OR
+            at least sample_title, dataset_ids, expression_assay_used, and sample_type should be populated
+
+  HighThroughputExpressionDatasetID:
```
markquintontulloch, line 203: Currently this is only used for `sample_id` in `HighThroughputExpressionDatasetSampleAnnotation`.  If this is not intended to be used for `dataset_id` then can it be written as a mixin or just incorporated into the class itself to avoid unnecessary join tables?

model/schema/highThroughputExpression.yaml
chris-grove: @markquintontulloch Yeah, so I debated with myself on this one. Generally, these are IDs for 3rd party databases, like GEO and ArrayExpress. Here's the slot definition description: 

"The ID from the data provider for the dataset. When available, this is the GEO ID. When no GEO ID is available, it will be the ArrayExpress ID. When these both are unavailable, it will be the MOD ID."

so, occasionally it will be a MOD identifier, but usually not. Do we really want to use 'mod_entity_id' for these? Or maybe we need to repurpose mod_entity_id to be db_entity_id to apply more generally? I just feel a bit odd trying to call these MOD IDs

model/schema/highThroughputExpression.yaml
chris-grove: @markquintontulloch Thanks for catching that. I will fix

model/schema/highThroughputExpression.yaml
markquintontulloch: I'd be in favour of renaming and repurposing `mod_entity_id` to be more general - I think we're already kind of stretching the definition in places already.  The other option here would be to inherit from `CurieObject` instead of `SubmittedObject` and add `data_provider` and `database_id` slots to this class.  As it is, it feels like `mod_entity_id` and `mod_internal_id` are never going to be used for this class, unless that's likely to change once we eventually move to direct submissions.

model/schema/highThroughputExpression.yaml
chris-grove: Maybe? I would have to ask curators if this would ever apply

model/schema/highThroughputExpression.yaml
chris-grove: Yes, fixing now

model/schema/highThroughputExpression.yaml
chris-grove: Maybe? I'd have to think on why there was a separate class created in the old JSON and whether having a separate class can be avoided. I've reworked the dataset_id slot. I may have missed your point

markquintontulloch: My point is moot given that this class is now being used for dataset_id and sample_id.  Having said that, I'm not sure that the modelling of this is the best way forward.
To me it seems like a HighThroughputExpressionDatasetAnnotation needs an ID field and cross-references (if we want to indicate a preferred crossreference then that could be achieved by an additional boolean slot on the CrossReference class).
Similary the HighThroughputExpressionDatasetSampleAnnotation needs an ID field and cross-references.  
Then we need to link the dataset samples to the datasets.  In the old schema this is done via the datasetIDs field on the sample entry but in LinkML this could be modelled as either an association, a slot that is a list of annotations on the sample class, or a slot that is a list of samples on the annotation class.
This would remove the need for this separate class (which if we were to keep should be an AuditedObject not a SubmittedObject) as the ID would be covered by modEntityId/modInternalId and each class would just have a cross_references slot.

model/schema/highThroughputExpression.yaml
```diff
@@ -0,0 +1,577 @@
...
+    domain: HighThroughputExpressionDatasetSampleAnnotation
+    range: sequencing_format_values
+
+  stage:
+    description: >-
+      The life stage of an organism or biosample as expressed by a TemporalContext entity
+    domain: BioSampleAge
+    range: TemporalContext
+
+  sub_series:
```
markquintontulloch, line 530: @chris-grove This really needs to be an association that is loaded after the datasets, otherwise we are reliant on the annotations being in the correct order when ingested.

chris-grove: @markquintontulloch  I realized that the range of sub_series should just be list of IDs, as strings I suppose. The entities being referred to in the list are not represented in our database, only in GEO, ArrayExpress, or in some cases, a MOD. I will make the appropriate change to make this a list of strings.

chris-grove:
@markquintontulloch @VarunReddy1111 
I think this should work but we should discuss more if not. Here are the significant changes:

- mod_entity_id replaced with primary_external_id
- sample* slots changed to htp_expression_sample*
- Changed HTPExpressionDataset to ExternalDatabaseEntity (nothing specific about the model just curie, secondary IDs, cross reference and preferred cross reference; could be anything)
- ExternalDataBaseEntity is_a CurieObject
- ExternalDataBaseEntityDTO is_a AuditedObjectDTO ('curie' slot declared)

markquintontulloch (approved):

PR #259, opened 2024-07-10
Assembly component tweak

markquintontulloch:
Switch display_name to name to bring in line with other GenomicEntity classes

### DIFF
--- a/model/schema/core.yaml
+++ b/model/schema/core.yaml
@@ -476,24 +476,27 @@ classes:
       In the context of this model, this entity is used as a landmark for the
       location of other genomic entities.
     slots:
-      - display_name
-      - genome_assembly
-    slot_usage:
-      curie:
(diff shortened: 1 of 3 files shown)


gildossantos (approved):

model/schema/core.yaml
```diff
@@ -476,24 +476,27 @@ classes:
       In the context of this model, this entity is used as a landmark for the
       location of other genomic entities.
     slots:
-      - display_name
+      - name
       - genome_assembly
+      - maps_to_chromosome
     slot_usage:
-      curie:
+      name:
```
sierra-moxon, line 483: just making sure you wanted to remove the curie field requirement here?

markquintontulloch: yes, curie is reserved for AGR curies and not currently a required field

PR #261, opened 2024-08-14
SCRUM-4300 add expression ribbon fields

adamgibs:


model/schema/expression.yaml
```diff
@@ -511,6 +513,18 @@ slots:
     range: GOTerm
     multivalued: false
 
+  cellular_component_ribbon_term:
+    description: >-
+      The primary cellular component slim term referred to within the context of an AnatomicalSite object and used by the expression ribbon.
+    domain: AnatomicalSite
```
oblodgett, line 519: @adamgibs shouldn't this be a GOTerm not an AnatomicalSite?

markquintontulloch: `domain: AnatomicalSite` is correct but it does need `range: GOTerm` too

adamgibs: Okay, I've just updated it.

markquintontulloch (approved):

model/schema/expression.yaml
```diff
@@ -511,6 +513,20 @@ slots:
...
+    range: GOTerm
+    multivalued: false
+
+  cellular_component_other:
+    description: >-
+      Is set to true when no cellular_component_ribbon_term is available
+    domain: AnatomicalSite
+    range: boolean
```
oblodgett, line 527: Is this range correct?

adamgibs: It matches other boolean values in the model like [here](https://github.dev/alliance-genome/agr_curation_schema/blob/b25a1dfb8f019d69d571664ec4b6e10dd877b8f9/model/schema/allele.yaml#L1024) so I think so.

PR #262, opened 2024-09-04
Added dataprovider to HTP linkMl model

VarunReddy1111:


chris-grove (changes requested):
@VarunReddy1111 
We should be reusing the already existing 'data_provider' slot name that has been re-used in other classes as well. Therefore we don't need to define a new slot. Actually, the only change that is needed here (I think) is the addition of the slot "data_provider" to the list of slots for HTPExpressionDatasetAnnotation as on (new) line 55 of highThroughputExpression.yaml. If that builds the JSON schema without error and runs tests without error, we can call it good.

chris-grove (approved):

PR #266, opened 2024-10-08
SCRUM-4441 Rewrite LinkML model for VEPGENE and VEPTRANSCRIPT loader data types

markquintontulloch:
Moves away from the "association of an association" model to just having a single class for VEP consequences that is attached to a CuratedVariantGenomicLocationAssociation (multivalued).

The gene-level consequence is simply the most severe transcript-level consequence for a variant amongst all the transcripts associated with a gene, so this can simply be indicated by a boolean rather than having separate gene- and transcript-level consequence classes.

Position slots prefixed with 'calculated_' to clearly indicate that these coordinates are determined by the VEP based on input files as opposed to source/curated values.

### DIFF
--- a/generated/jsonschema/allianceModel.schema.json
+++ b/generated/jsonschema/allianceModel.schema.json
@@ -8879,6 +8879,13 @@
                     "description": "A base preceded an insertion or deletion event that is reported when either the reference or variant sequence would otherwise be empty. ",
                     "type": "string"
                 },
+                "predicted_variant_consequences": {
+                    "description": "VEP-calculated variant consequences",
+                    "items": {
+                        "$ref": "#/$defs/PredictedVariantConsequence"
+                    },
+                    "type": "array"
+                },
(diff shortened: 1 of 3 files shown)


sierra-moxon (approved):

chris-grove:
@markquintontulloch Looks good and I generally approve; one question, do we have the deleted enumerations noted elsewhere? I don't want to lose this information if this is the only place it has been noted.
```
enums:

  vep_consequence_levels:
    permissible_values:
      high:
      moderate:
      low:
      modifier:

  sift_prediction_levels:
    permissible_values:
      deleterious:
      tolerated:

  polyphen_prediction_levels:
    permissible_values:
      possibly_damaging:
      probably_damaging:
      benign:
```

markquintontulloch:
@chris-grove permissible values listed in attributes

chris-grove (approved):
@markquintontulloch  OK, got it, thanks

PR #270, opened 2024-10-23
Scrum 4190

cmpich:


chris-grove:
@cmpich  I'm a bit concerned that these additions don't really conform with the existing LinkML standards that we've been using. What has been previously agreed is that every object should inherit from some base class, at least AuditedObject or AuditedObjectDTO. Given that this is an annotation class, I would think we would want to inherit from Annotation as "GeneOntologyAnnotation". Something like this:

```
classes: 

    GeneOntologyAnnotation:
      is_a: Annotation
      description: An annotation associating a gene to a Gene Ontology (GO) term
      slots:
        - gene_ontology_annotation_subject
        - gene_ontology_annotation_object

    GeneOntologyAnnotationDTO:
      is_a: AnnotationDTO
      description: An annotation associating a gene to a Gene Ontology (GO) term (DTO class)
      slots:
        - gene_identifier
        - go_term_curie
      slot_usage:
        gene_identifier:
          required: true

slots: 

    gene_ontology_annotation_subject:
      description: the gene to which the Gene Ontology (GO) term is being annotated
      domain: GeneOntologyAnnotation
      range: Gene
      mulivalued: false
      required: true
    
    gene_ontology_annotation_object:
      description: the Gene Ontology (GO) term to which the gene is being annotated
      domain: GeneOntologyAnnotation
      range: GOTerm
      mulivalued: false
      required: true

    go_term_curie:
      description: the Gene Ontology (GO) term to which the gene is being annotated
      domain: GeneOntologyAnnotationDTO
      range: string
      mulivalued: false
      required: true
```

@markquintontulloch  What are your thoughts?

chris-grove:
One more note:
If we do inherit from Annotation, then we would be required to include values for the following fields:

```
data_provider
single_reference
relation
internal
```

The top three are required in the GAF file and internal can default to false

chris-grove (commented):
See above comments. I'd like to discuss a bit before approving

markquintontulloch:
I agree with @chris-grove

markquintontulloch:
Although perhaps it would not be appropriate to inherit from Annotation since I'm guessing conditionRelations and relatedNotes would not apply?

chris-grove:
@markquintontulloch 
It's true that these annotations do not currently have notes or conditions, but these are not required fields and so I think it is not problematic that these GO annotations will not make use of these fields. OR we could push these fields down into the leaf nodes that need these fields and omit them from the ancestral classes

markquintontulloch:
@chris-grove I agree that it won't be problematic as the fields aren't required, but if they will never be populated then is it worth having them for the class?  Just because something is an "annotation" does that necessitate that it inherits from the Annotation class?  It could just inherit from AuditedObject and we could just add the fields that will actually be used.  I don't have any strong feelings either way, but there's a few ways this could be achieved.

Having said all of that, this PR is just for a DTO class.  There is no reason the DTO class needs to inherit from anything, it just needs to match the input source and these are coming in from URLs not submitted files.  In that sense, maybe this PR is fine but we need to add the non-DTO class.

chris-grove:
I just fixed my "mulivalued" typos

chris-grove (approved):

PR #274, opened 2024-12-03
Add slots to AGM class and test file

chris-grove:
New slots: synonyms, parental_populations

### DIFF
--- a/model/schema/affectedGenomicModel.yaml
+++ b/model/schema/affectedGenomicModel.yaml
@@ -42,6 +42,7 @@ classes:
     slots:
       - name
       - subtype
+      - synonyms
       - agm_secondary_ids
       - components
       - sequence_targeting_reagents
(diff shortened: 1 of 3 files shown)


model/schema/affectedGenomicModel.yaml
```diff
@@ -182,7 +185,12 @@ slots:
   parental_populations:
```
sierra-moxon, line 185: Do we still need this slot with this change?

chris-grove: @sierra-moxon 
The parental_population slot is for the record in the database, the parental_population_curies slot is only for the DTO class for submitted AGMs. They are both needed, I think.

model/schema/affectedGenomicModel.yaml
```diff
@@ -182,7 +185,12 @@ slots:
...
     singular_name: parental_population
     domain: AffectedGenomicModel
-    range: uriorcurie
+    range: AffectedGenomicModel
+
+  parental_population_curies:
+    domain: AffectedGenomicModelDTO
```
markquintontulloch, line 191: Reference to an AGM from an AGM is going to enforce loading order constraints (or a two-stage load process).  I guess this is okay, an Association and separate load seems overkill.  Possibly worth discussing with the rest of the A-Team.

PR #279, opened 2025-01-06
Adding agmAlleleAssociations object to AGM

VarunReddy1111:


markquintontulloch (changes requested):
This change isn't necessary.  The components slot is the AGMAlleleAssociations

PR #287, opened 2025-03-31
recognize casing of Ingest class

sierra-moxon:
WARNING:linkml.generators.jsonschemagen:No class in schema named ingest
cp -pr generated/target/jsonschema generated/

better warning message in more recent versions of linkml notified us that we need to update the casing of "Ingest" class in order to be recognized as a valid class in agr_curation_schema.


markquintontulloch (approved):

chris-grove (approved):
Works, thanks!

PR #292, opened 2025-05-30
Cassette class addition

chris-grove:


chris-grove:
Note that this PR is almost entirely making a carbon copy of the Construct class and renaming Construct/construct to Cassette/cassette

chris-grove:
Sorry, I was a bit too hasty on that last commit. I will clean it up

chris-grove:
@sierra-moxon 
I've added a new slot definition for `placeholder` in the core.yaml file and added the slot to the Reagent superclass

PR #295, opened 2025-08-02
Fix allele construct association relation example

chris-grove:
and add examples of all current valid relations


JoelRichardson (approved):
Thanks @chris-grove. I think this looks good.

PR #297, opened 2025-08-22
Adding boolean terms to flag the Other categorized  Uberon Terms

VarunReddy1111:
Added 2 boolean variables, which will be changed to true if the Uberon term which is categorized as Other is submitted


chris-grove (approved):
@VarunReddy1111 Approved, thanks!

PR #322, opened 2026-04-28
Fix functional gene set test

chris-grove:
@sjm41 Super simple fix needed on the functional gene set test JSON file. There was an unexpected trailing comma that I removed.


sjm41 (approved):
Wonder how that got there??

PR #323, opened 2026-05-14
Add GeneChangeEventSlotAnnotationDTO

sjm41:
## Summary
- Adds `GeneChangeEventSlotAnnotationDTO` (mirrors `AlleleNomenclatureEventSlotAnnotationDTO` precedent) so gene change events become ingestable.
- Wires `gene_change_event_dtos` into `GeneDTO`.
- Adds DTO-side `event_type_name` / `event_status_name` slots (VocabularyTerm references reified as strings).
- Reuses primitive-typed slots (`current_version`, `symbol_renamed_to/from`, `full_name_renamed_to/from`) across persistence and DTO classes; their persistence-only `domain:` declarations have been dropped so the slots are accurately class-agnostic, matching the `core.yaml` shared-slot pattern.

## Deferred — known gap
Merge/split fields (`acquires_in_merge`, `merged_into`, `split_from`, `split_into`) are intentionally **not** mapped into the DTO. The persistence-side slots have no explicit `range:` declared, so the DTO mapping is ambiguous (presumably gene-identifier strings, but worth confirming). Best resolved in a follow-up PR that also fixes the persistence-side range gap.

## Test plan
- [x] CI (`check-pull-request.yaml`) regenerates artifacts and runs the full test suite — local rebuild was skipped (no poetry on dev machine)
- [x] Reviewer sanity-checks that the new DTO surfaces as expected in the regenerated `allianceModel.schema.json`
- [x] Reviewer confirms the deferred merge/split decision is acceptable


chris-grove (changes requested):
@sjm41 OK, it generally looks good but I think we should consider a modified approach. Here's what I suggest:
1. We create a generic parent class ChangeEventSlotAnnotation which has all the same properties, except single_gene is excluded; this generic class can be the parent class for change events for all other types of entities for which we want to track change events. This parent class should be considered abstract so "abstract: true" should be declared in the parent class definition.
2. The range of 'acquires_in_merge', 'merged_into', 'split_into' and 'split_from' will remain undefined in the slot definitions and the parent class declaration
3. Make GeneChangeEventSlotAnnotation a child of the generic ChangeEventSlotAnnotation with the class definition introducing (only) the 'single_gene' slot (this should be declared as required in slot_usage). In addition, this class definition should include slot usage to declare that the range of the slots 'acquires_in_merge', 'merged_into', 'split_into' and 'split_from' is a Gene
4. Once those upper level changes are in place, recreate the DTO classes such that a ChangeEventSlotAnnotationDTO class has the the slots 'acquires_in_merge_curies', 'merged_into_curie', 'split_into_curies' and 'split_from_curie' which will also need their own slot definitions (each with range 'string'). The slots 'acquires_in_merge', 'acquires_in_merge_curies', 'split_into' and 'split_into_curies' should all be declared as multivalued.
5. Create a child of ChangeEventSlotAnnotationDTO that is GeneChangeEventSlotAnnotationDTO that declares the slot 'gene_identifier' (reused)

sjm41:
Thanks @chris-grove !
I believe Claude has actioned all of your requests - see https://github.com/alliance-genome/agr_curation_schema/pull/323/changes/dd7b84f69a4c7d8c8807df81fbcba961cbbdde32

Note that 2 additional changes were suggested, which I accepted:
- Moved the relevant slot definitions from gene.yaml into core.yaml
  alongside the new parent classes (matching the NameSlotAnnotation
  precedent), and dropped class-specific domain: declarations so the
  slots are reusable.
- Corrected split_from to multivalued: false to match its single-source
  semantics and the new split_from_curie DTO naming.

Sound OK to you?

chris-grove:
@sjm41 Sounds good, I'll have a look!

chris-grove (approved):
Looks good!

model/schema/gene.yaml
```diff
@@ -180,43 +181,39 @@ classes:
...
-      the split get deprecation type ChangeEvent annotations.
-      deprecation - The entity was made obsolete. This may include objects
-      deprecated by some merge/split event.
-      rename - A current accepted name of the entity was changed.
-      type_change - The entity type was changed.
-      Proposed status vocabulary: provisional/approved.
+    is_a: ChangeEventSlotAnnotation
+    description: >-
+      A major change to a gene: e.g., new, merge, split, rename, deprecation,
+      type_change.
```
sierra-moxon: Is `type_change` really supported here?  If so, should we record type-from-->type-to sorts of changes?  And if so, would we still call this a "Gene"ChangeEvent?

ianlongden:
Can I request that all PR's have the Jira ticket number as part of the title so that the ticket can be easily found. Thanks. It just makes it much easier to track.

model/schema/gene.yaml
chris-grove: @sierra-moxon Good question. I will defer to @sjm41 and how they do things at FlyBase, which is where this model is largely derived from

chris-grove:
> Can I request that all PR's have the Jira ticket number as part of the title so that the ticket can be easily found. Thanks. It just makes it much easier to track.

@ianlongden I don't believe there is any associated Jira ticket. Not all model changes have a Jira ticket associated

sjm41:
> > Can I request that all PR's have the Jira ticket number as part of the title so that the ticket can be easily found. Thanks. It just makes it much easier to track.
> 
> @ianlongden I don't believe there is any associated Jira ticket. Not all model changes have a Jira ticket associated

Right, there's no (current) Alliance JIRA ticket related to this - the fields being modified were originally created some time ago. We do have an FlyBase JIRA ticket for submitting these data, and I did quote this GitHub issue on there to try to help with tracking.

model/schema/gene.yaml
sjm41: Thanks @sierra-moxon. I guess this is meant to refer to biotype changes, e.g. protein-coding -> lncRNA? If so, I think this could be validly classed as a 'gene change event'. However, this is not something FlyBase records/tracks in the same way as  merges, splits, renames, so I'll remove this part of the description, at least for now. (we can consider adding it properly later if someone needs it.)

PR #325, opened 2026-06-03
SCRUM-6110: Ltp expression dto classes and properties

chris-grove:
## Summary

Adds the ingest path for LTP gene-expression submissions, and refactors the expression entity model so that the gene and its supporting reference are stated once per experiment rather than repeated on every annotation.

## Entity model changes

- **`ExpressionAnnotation` now inherits from `Association` instead of `Annotation`.** The annotation no longer carries its own subject or reference — `expression_annotation_subject`, the `single_reference` requirement, and `data_provider` are removed, since all three are properties of the parent `GeneExpressionExperiment`. Slots still wanted on the annotation (`curie`, `unique_id`, `primary_external_id`, `mod_internal_id`, `related_notes`) are declared directly, as they were previously inherited via `Annotation`.
- **`GeneExpressionExperiment.single_reference` is now required** — every experiment must cite a reference.
- **`when_expressed_stage_name` and `where_expressed_statement` are reclassified as derived values.** They remain required on the entity, but are *not submitted*: the loading pipeline assembles them after ingest from the controlled terms of the `expression_pattern` together with the submitted free-text values. Both are redescribed as human-readable summaries composed of ontology term labels and/or free text, and both are absent from the DTOs.
- **New free-text slots `when_expressed_free_text` and `where_expressed_free_text`** on `ExpressionAnnotation`, for the cases where controlled terms can't fully capture the pattern. These are what submitters provide, and they feed the post-ingest assembly above.
- **Fix:** `AnatomicalSiteDTO` inherited from `AuditedObject`; corrected to `AuditedObjectDTO`. Its `created_by`/`updated_by` now correctly generate as `created_by_curie`/`updated_by_curie`.
- **`AnatomicalSiteDTO` filled out to mirror `AnatomicalSite`:** adds `cellular_component_ribbon_term_curies` (new slot) plus the existing booleans `cellular_component_other`, `anatomical_structure_uberon_term_other`, and `anatomical_substructure_uberon_term_other`. The `domain:` was dropped from those three booleans so they are reusable by both the entity and the DTO under the same name.
- The FMS→LinkML mapping notes on `GeneExpressionAnnotation` are updated to reflect that `geneId` and `evidence` now map to `GeneExpressionExperiment`, and that `stageName`/`whereExpressedStatement` are assembled post-ingest.

## New DTO classes

- **`GeneExpressionExperimentDTO`** (`SubmittedObjectDTO`) — the top-level submission class. Requires `gene_identifier`, `reference_curie`, and `expression_assay_curie`; also carries detection reagents, specimen genomic model / alleles, condition relations, and cross references. Child annotations are inlined via `expression_annotation_dtos` so submitters never back-reference the experiment by id.
- **`GeneExpressionAnnotationDTO`** (`AuditedObjectDTO`) — one annotation. Requires `expression_pattern_dto` and `relation_name`; accepts `when_expressed_free_text` / `where_expressed_free_text`, negation/uncertainty flags, figure identifiers, cross references, and notes. Deliberately has no gene, no reference, and no stage-name/where-statement.
- **`ExpressionPatternDTO`** — pairs `when_expressed_dto` (`TemporalContextDTO`, optional) with `where_expressed_dto` (`AnatomicalSiteDTO`, required).

Supporting slots added: `expression_pattern_dto`, `when_expressed_dto`, `where_expressed_dto`, `detection_reagent_identifiers`, `specimen_genomic_model_identifier`, `specimen_allele_identifiers`, `related_figure_identifiers`, `expression_annotation_dtos`, `cellular_component_ribbon_term_curies`, `when_expressed_free_text`, `where_expressed_free_text`.

## Ingest wiring

- `gene_expression_experiment_ingest_set` registered on the `Ingest` class, and `expression` added to `ingest.yaml` imports.

## Tests

- `test/data/gene_expression_test.json` — valid fixture: one experiment with two annotations, exercising both the anatomical-structure and cellular-component branches of the `AnatomicalSiteDTO` postcondition, plus free text at the annotation level.
- `test/data/invalid/gene_expression_invalid.json` — correctly rejected on three counts: missing `gene_identifier`, missing `expression_assay_curie`, and missing `where_expressed_dto`.
- Both registered in the `Makefile`.
- `generated/jsonschema/allianceModel.schema.json` regenerated.

## Review feedback addressed

- **Image associations** are intentionally absent from the annotation DTO; a DTO for expression-annotation-image-associations will follow once there are images to submit.
- **Gene/reference duplication between parent and child** — resolved by the `Association` refactor above; they are now stated once, on the experiment.
- **`where_expressed_free_text` / `when_expressed_free_text` placement** — these were initially added to `AnatomicalSite` and `TemporalContext`, but neither class is expression-specific: `AnatomicalSite` is also the range of `htp_expression_sample_locations`, and `TemporalContext` of `BioSampleAge.stage`. Both slots were moved up to `ExpressionAnnotation` / `GeneExpressionAnnotationDTO`, leaving the two reusable classes generic rather than renaming them.
- **Stage name vs. stage identifier** — `when_expressed_stage_name` is no longer submitted at all, so it no longer competes with the stage term id. The invalid fixture's `"Prim-5"` (an actual ZFS term label, which made the field look redundant) is replaced with genuine free text, `"expressed prior to hatching"`.

Note that `when_expressed_stage_name` is a shared slot and `BioSampleAgeDTO` *does* accept it on submission, so the "not submitted" note on the slot definition is scoped to `ExpressionAnnotation` rather than asserted globally.

## Test plan

- [x] `make test-jsonschema` — 34/34 valid fixtures pass, including `gene_expression_test.json`
- [x] `make test-jsonschema_invalid` — 5/5 invalid fixtures correctly rejected, including `gene_expression_invalid.json`
- [ ] Confirm downstream consumers (curation-api) can deserialize the new DTOs


chris-grove:
Responding to the questions from Claude Review:
1. We will later establish a DTO for expression-annotation-image-associations once we have images to submit; not needed right now
2. This has been fixed by refactoring the ExpressionAnnotation class to inherit directly from Association instead of Annotation and clearing out the unnecessary and redundant properties
3. Also addressed with the refactor mentioned in number 2

chris-grove:
Just added one change to make single_reference of ExpressionExperiment required

model/schema/expression.yaml
```diff
@@ -297,6 +297,7 @@ classes:
       - cellular_component_qualifiers
       - anatomical_structure_uberon_term_other
       - anatomical_substructure_uberon_term_other
+      - where_expressed_free_text
```
sierra-moxon: From an outside perspective, this seems a bit misplaced.  The class describes an anatomical site (in a specimen), but then we talk about "expression" without any reference to the gene, etc.  And in fact, we have other classes that describe expression gene, stage, etc.  I get that this might be a convenience placement w/re to class hierarchy.

sierra-moxon: If AnatomicalSite is only used to describe where expression is located, then changing the name of this class (and/or its definition) to convey that would clear up my comment here.

model/schema/expression.yaml
```diff
@@ -279,6 +278,7 @@ classes:
       - age
       - temporal_qualifier_names
       - stage_uberon_slim_term_curies
+      - when_expressed_free_text
```
sierra-moxon: its not clear why "when_expressed_stage_name" is also not in this class.  Same comment about "expressed" keyword in this class.  If only used for expression, the class name and/or description should change too.  otherwise this expression field and (when_expressed_stage_name field above) feel out of place.

test/data/invalid/gene_expression_invalid.json
```diff
@@ -0,0 +1,35 @@
...
+      "data_provider_dto": {
+        "source_organization_abbreviation": "ZFIN",
+        "internal": false
+      },
+      "created_by_curie": "ZFIN",
+      "updated_by_curie": "ZFIN",
+      "internal": false,
+      "expression_annotation_dtos": [
+        {
+          "when_expressed_stage_name": "Prim-5",
```
sierra-moxon: this example leaves me wondering why we need stage name when we have a stage identifier.  maybe we need an example where stages aren't ontology terms to make this more clear?  

If that is the use case for this field, maybe we need to consider refactoring the "required" slots on when_expressed_dto to be inclusive of this use case and able to be used consistently vs. the free text at the upper object level?

model/schema/expression.yaml
chris-grove: Agreed, and you're right that this isn't just a naming nit — `AnatomicalSite` genuinely isn't expression-specific. It's also the range of `htp_expression_sample_locations` ("the location within an organism from which the sample was taken"), so a `where_expressed_*` slot on it doesn't apply there.

Fixed by moving the slot rather than renaming the class: `where_expressed_free_text` now lives on `ExpressionAnnotation` / `GeneExpressionAnnotationDTO`, where "expressed" is meaningful, and `AnatomicalSite`/`AnatomicalSiteDTO` are back to being purely a set of anatomical terms.

To your follow-up about renaming the class or its definition instead — that would have been the right move *if* `AnatomicalSite` were expression-only, but it isn't (see the HTP sample-location reuse above), so keeping the class generic and relocating the expression-specific slot is the better fit. No rename needed.

model/schema/expression.yaml
chris-grove: Same resolution as the `AnatomicalSite` thread: `when_expressed_free_text` has been moved off `TemporalContext`/`TemporalContextDTO` and onto `ExpressionAnnotation`/`GeneExpressionAnnotationDTO`. `TemporalContext` is reused by `BioSampleAge.stage` in the high-throughput expression model, so it shouldn't carry expression-specific slots either.

On why `when_expressed_stage_name` isn't in `TemporalContext` — it stays at the annotation level deliberately, and this PR now makes the reason explicit. It is *not submitted*: the loading pipeline assembles it after ingest from the `TemporalContext` terms of the `expression_pattern` **plus** the submitted `when_expressed_free_text`. It's a human-readable summary spanning both, so it belongs on the annotation rather than inside either component. Same treatment for `where_expressed_statement`.

Both are consequently removed from `GeneExpressionAnnotationDTO` entirely — submitters no longer supply them.

One caveat worth noting: `when_expressed_stage_name` is a shared slot, and `BioSampleAgeDTO` in `highThroughputExpression.yaml` *does* accept it on submission. So the "not submitted" note on the slot definition is scoped to `ExpressionAnnotation` rather than asserted globally.

test/data/invalid/gene_expression_invalid.json
chris-grove: Good catch — "Prim-5" was a poor choice precisely because it *is* an existing ZFS ontology term label, which made the field look redundant with the stage identifier. That example was misleading about the intent.

Resolved on both fronts:

1. `when_expressed_stage_name` is gone from `GeneExpressionAnnotationDTO`. It isn't submitted at all; it's assembled post-ingest from the ontology terms in `when_expressed_dto` together with the submitted `when_expressed_free_text`. So there's no longer a stage name *and* a stage identifier competing at submission time.
2. The fixture now reads `"when_expressed_free_text": "expressed prior to hatching"` — genuine free text that no ontology term label covers, which is the actual use case for the field.

On refactoring the required slots of `when_expressed_dto`: `TemporalContextDTO` has no required slots, so a submitter with no usable stage term can omit `when_expressed_dto` and supply only `when_expressed_free_text` at the annotation level. `where_expressed_dto` remains required on `ExpressionPatternDTO` with its existing `anatomical_structure_curie`-or-`cellular_component_curie` postcondition, matching the FMS requirement. If the expression WG wants a fully term-free spatial submission path as well, that's a larger change — happy to open a separate ticket if you think it's needed.

model/schema/expression.yaml
chris-grove: Comments above are Claude-generated but I agree with the sentiment. Another point is that the new AffectedEntity class for PhenotypeAnnotation will make use of AnatomicalSite and TemporalContext in an a phenotype context, also not expression related.

chris-grove:
@sierra-moxon With the help of Claude, I think I've addressed all of your comments. Your review would be greatly appreciated. Thanks!

PR #326, opened 2026-06-04
Make event_status and current_version optional on gene change events

sjm41:
## Summary
- Follow-up to #323. Overrides `required: false` via `slot_usage` on both `GeneChangeEventSlotAnnotation` and `GeneChangeEventSlotAnnotationDTO` for `event_status` / `event_status_name` and `current_version`.
- Slots remain `required: true` on the abstract `ChangeEventSlotAnnotation` / `ChangeEventSlotAnnotationDTO` parents — only the gene subclass relaxes them.

## Test plan
- [ ] CI (`check-pull-request.yaml`) regenerates artifacts and runs the full test suite


sjm41:
Sorry, meant to make these changes on the parent classes, not just the gene-specific children. Now fixed as per https://github.com/alliance-genome/agr_curation_schema/pull/326/changes/14f9ee079df12ede374bccb81808768ad91500a9

chris-grove (approved):
I also went ahead and removed the slot requirements at the level of the slot definitions, which is required in this case as we cannot override requirements for slots in slot_usage if they are defined as required in the slot definitions. I've made those changes and all tests pass so I'm approving this PR

PR #327, opened 2026-07-02
Add optional is_aberration and is_balancer flags to Allele

sjm41:
## Summary

FlyBase makes a clear distinction between traditional single-gene alleles and chromosomal aberrations (multi-gene alleles) — handling them as separate data classes (FBal vs FBab) — and balancers are a subtype of aberration. This adds two optional booleans to `Allele`/`AlleleDTO` to capture that distinction within the Alliance single `Allele` class:

- **`is_aberration`** — indicates the allele is a chromosomal aberration (multi-gene allele) rather than a traditional single-gene allele.
- **`is_balancer`** — indicates the allele is a balancer (a recombination-suppressing chromosomal rearrangement). Balancers are a subtype of aberration, so this is only applicable to alleles for which `is_aberration` is true (documented in the slot description; not schema-enforced).

Both follow the existing `is_extinct` / `is_extrachromosomal` / `is_integrated` pattern — plain optional booleans reused directly in the DTO.

## Also

Removes the dead `aberration` `slot_usage` entry (flagged by `[CG]` as no longer declared) and its unused standalone slot definition, which `is_aberration` supersedes. Both were already absent from the generated artifacts.

## Notes

- No LinkML `rule` enforcing balancer ⟹ aberration (considered, deliberately left out).
- `generated/` artifacts will be regenerated by CI.

### DIFF
--- a/generated/jsonschema/allianceModel.schema.json
+++ b/generated/jsonschema/allianceModel.schema.json
@@ -3309,7 +3309,7 @@
                     "type": "boolean"
                 },
                 "is_aberration": {
-                    "description": "Indicates whether the allele is a chromosomal aberration (a multi-gene allele), as opposed to a traditional single-gene allele.",
+                    "description": "Indicates whether the allele is a chromosomal aberration, defined (per SO:1000183 - chromosome_structure_variation) as an alteration of the genome that leads to a change in the structure or number of one or more chromosomes.",
                     "type": [
                         "boolean",
                         "null"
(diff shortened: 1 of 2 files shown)


sjm41:
FYI: I put this proposal on the agenda for the Disease and Pheno WG meeting on July 8th.

gm119 (changes requested):
I would change the 'is_aberration' descriptions so that it avoids mention of 'a multi-gene allele' as a single transgenic insertion can also affect multiple genes (e.g. insertion into overlapping genes), and we would not regard those as an 'aberration'.

Our current FB QA/QC checks require us to type aberrations with a SO term that sits in the 'chromosome_structure_variation ; SO:1000183' branch, this term is defined as 'An alteration of the genome that leads to a change in the structure or number of one or more chromosomes.'

So I think the key part is the  'change in the structure or number of one or more chromosomes'

So I would use the SO term as the basis of the linkml definition for is_aberration instead of having 'a multi-gene allele'., so maybe:

  is_aberration:
    description: >-
      Indicates whether the allele is a chromosomal aberration, defined as 'An alteration of the genome that leads to a change in the structure or number of one or more chromosomes.' (SO:1000183).
    domain: Allele
    range: boolean

(or however SO terms should be referred to)

gm119 (approved):
looks good

chris-grove (approved):

PR #330, opened 2026-09-02
SCRUM-6283 Add ChromosomeAccession class

VarunReddy1111:
## Summary
Adds a `ChromosomeAccession` class to `core.yaml`, modelling the chromosome-name / RefSeq-accession / genome-assembly reference data that is currently hard-coded in the `ChromosomeAccessionEnum` in `agr_curation`. Moving this into the schema (and a DB table on the app side) lets new genome assemblies be added via data load instead of a code change.

Companion to agr_curation PR #2869 (SCRUM-6283), which adds the JPA entity, DAO, service, and Flyway-seeded table.

## Details
- `is_a: AuditedObject`, placed alongside `Chromosome` / `AssemblyComponent` in `core.yaml`.
- Fields defined as inline `attributes` (all `required`): `chromosome_name`, `accession`, `assembly_identifier`, `display_order`.
- `display_order` preserves the enum's original declaration order so accessions shared across assemblies (e.g. `RefSeq:NC_002333.2` for GRCz11 & GRCz12tu) resolve to the earlier-declared assembly, matching prior behaviour.
- **No `...DTO` class** — this is internal reference data seeded via migration with no ingest path (mirrors `OntologyTermClosure`).

## Notes
- No `generated/` edits (CI regenerates artifacts).
- No new test data added: this is a new class within an existing domain (`core.yaml`), not a new domain, and it has no ingest JSON-schema surface to exercise.

### DIFF
--- a/generated/jsonschema/allianceModel.schema.json
+++ b/generated/jsonschema/allianceModel.schema.json
@@ -14426,6 +14426,94 @@
             "title": "Chromosome",
             "type": "object"
         },
+        "ChromosomeAccession": {
+            "additionalProperties": false,
+            "description": "A mapping of a chromosome name to its RefSeq accession within a specific genome assembly. Holds the reference data formerly hard-coded in the ChromosomeAccessionEnum so that new genome assemblies can be added via data load rather than a code change.",
+            "properties": {
+                "accession": {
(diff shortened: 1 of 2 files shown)


chris-grove (approved):
I might have a few questions about this including the questions/points raised by Claude in the description, but it seems reasonable and all tests pass so I will approve

PR #332, opened 2026-09-09
SCRUM-6496 change Antibody taxon/antigen_taxon slots to VocabularyTerm

kthorat-prog:
## Summary
- `Antibody.taxon` / `Antibody.antigen_taxon` were `NCBITaxonTerm`-ranged, requiring a real NCBITaxon ID. Curators need to record free-text/not-specified values (e.g. "mouse", "not specified") alongside real NCBITaxon IDs, so both move to `VocabularyTerm`-ranged `taxon_term`/`antigen_taxon_term`.
- DTO slots renamed to match: `taxon_curie`/`antigen_taxon_curie` -> `taxon_term_string`/`antigen_taxon_term_string` (range stays `string`).
- The `_term_string` suffix (rather than the usual VocabularyTerm -> `_name` convention) is deliberate: unlike e.g. `clonality_name`, this field holds either an NCBITaxon curie or free text, not a human-readable name.
- The two backing controlled vocabularies (Antibody Taxon CV, Antibody Antigen Taxon CV) are being built on the `agr_curation` side from the non-redundant term lists curators provided, with the first column (including NCBITaxon IDs) as the vocabulary term "Name" and the second column (human readable) as "Definition" -- verified the `agr_curation` migration's term lists exactly match the source spreadsheet.
- Updated `test/data/antibody_test.json`'s example values to terms that exist in the new non-redundant lists.

## Test plan
- [x] Confirmed no other class references the renamed `antigen_taxon`/`antigen_taxon_curie` slots (they were `domain: Antibody`/`domain: AntibodyDTO` restricted already).
- [x] Confirmed the shared `taxon`/`taxon_curie` slots in `core.yaml` are untouched and still used correctly by other classes (`ConstructComponentSlotAnnotation`, etc.).
- [x] `antibody_test.json` is already registered in the Makefile's `SCHEMA_TEST_EXAMPLES`.

### DIFF
--- a/generated/jsonschema/allianceModel.schema.json
+++ b/generated/jsonschema/allianceModel.schema.json
@@ -9040,8 +9040,15 @@
                         "null"
                     ]
                 },
+                "antigen_taxon": {
+                    "description": "Holds between an Antibody and the species from which the antigen originates (as represented by taxon). Antigen source species is an open set (potentially any organism used to raise the antigen), unlike the small closed set of common antibody hosts, so unlike host_taxon_term this stays an NCBITaxonTerm reference rather than a closed CV -- see antigen_taxon_term for why a species is absent (not specified, etc.).",
+                    "type": [
+                        "string",
+                        "null"
+                    ]
+                },
                 "antigen_taxon_term": {
-                    "description": "Holds between an Antibody and the species from which the antigen originates. A VocabularyTerm rather than an NCBITaxonTerm so free-text/not-specified values can be recorded alongside real NCBITaxon IDs.",
+                    "description": "Records *why* the antigen's source species is absent (not specified, etc.) when antigen_taxon isn't submitted. This CV is disjoint from NCBITaxon and must never contain an NCBITaxon curie: a taxonomy-aware query (e.g. \"any antibody against a Drosophila antigen\") only needs to check antigen_taxon, and allowing curies here would reintroduce the ambiguity of two possible locations for a species value.",
                     "type": [
                         "string",
                         "null"
@@ -9276,8 +9283,15 @@
                         "null"
                     ]
                 },
+                "antigen_taxon_curie": {
+                    "description": "Curie of the NCBITaxonTerm representing the species from which the antigen originates (as represented by taxon).",
+                    "type": [
+                        "string",
+                        "null"
+                    ]
+                },
                 "antigen_taxon_term_name": {
-                    "description": "Name of the VocabularyTerm representing the species from which the antigen originates: either an NCBITaxon curie or a free-text/not-specified value. From the Antibody Antigen Taxon CV.",
+                    "description": "Name of the VocabularyTerm recording why the antigen's source species is absent: a free-text/not-specified value, never an NCBITaxon curie (submit antigen_taxon_curie instead when the species is known). From the Antibody Antigen Taxon CV.",
                     "type": [
                         "string",
                         "null"
--- a/generated/jsonschema/allianceModel.schema.json
+++ b/generated/jsonschema/allianceModel.schema.json
@@ -9041,7 +9041,7 @@
                     ]
                 },
                 "antigen_taxon_term": {
-                    "description": "Holds between an Antibody and the species from which the antigen originates (as represented by taxon). A VocabularyTerm rather than an NCBITaxonTerm so free-text/not-specified values can be recorded alongside real NCBITaxon IDs.",
+                    "description": "Holds between an Antibody and the species from which the antigen originates. A VocabularyTerm rather than an NCBITaxonTerm so free-text/not-specified values can be recorded alongside real NCBITaxon IDs.",
                     "type": [
                         "string",
                         "null"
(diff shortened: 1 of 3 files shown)


chris-grove:
@kthorat-prog  I'm looking at this now

chris-grove:
@kthorat-prog I've reviewed this and pushed the naming fixes directly to the branch in the latest two commits (`89662104`, `131d4dd0`) — CI is green, all 33 `SCHEMA_TEST_EXAMPLES` validate. Details below, plus a proposal on `antigen_taxon` I'd like your and curation's take on.

## What I changed

**DTO suffix `_string` → `_name`.** The rationale in the PR description doesn't quite hold. `VocabularyTerm.name` is `identifier: true` (`controlledVocabulary.yaml:83-85`), and the CV is being built with the curie-bearing column as the term **Name** — so the value submitted here *is* `VocabularyTerm.name`. The `_name` suffix denotes which field of the VocabularyTerm gets submitted, not whether the string happens to read as a human-readable label. Renamed to `antigen_taxon_term_name` (`:751`) and `host_taxon_term_name` (`:1009`), which also lines them up with `clonality_name` / `heavy_chain_isotype_name` / `light_chain_isotype_name` on the same class.

**Added a `host_` prefix.** `core.yaml:1280` `taxon` has no `domain:` and is used by 7 classes, so an unprefixed `taxon_term` sitting next to it in the global namespace is a real mix-up risk — the next class wanting a VocabularyTerm-ranged taxon would pick it up carrying antibody-specific semantics. Went with `host_taxon_term` rather than `antibody_taxon_term`: it states what the slot actually means ("the species in which the antibody was raised") and reads as a natural counterpart to `antigen_taxon_term`. Confirmed no collision anywhere in the model.

**Deduplicated the FB note.** It now lives only at `Antibody.slot_usage.host_taxon_term` (`:113`).

**Two extras while I was in there:** dropped the `(as represented by taxon)` parenthetical from both `antigen_taxon_term` descriptions (`:740`, `:751`) — inaccurate now that the range isn't a taxon reference; and moved the `host_taxon_term` definitions into alphabetical position in the `slots` block (`:998`), between `heavy_chain_isotype_name` and `light_chain_isotype`.

**Renamed the CV** to "Antibody Host Taxon CV" (`:1014`) to match the slot.

## ⚠️ Needed on the `agr_curation` side

That CV rename means the migration needs the same change — vocabulary `name` only, the term list is unaffected. It joins three items that were already outstanding:

1. Vocabulary rename: `Antibody Taxon CV` → `Antibody Host Taxon CV`
2. Loader field mapping: `taxon_curie` → `host_taxon_term_name`, `antigen_taxon_curie` → `antigen_taxon_term_name`
3. `NCBITaxonTerm` → `VocabularyTerm` FK change on both `Antibody` columns
4. Data migration for existing `Antibody` rows currently pointing at `NCBITaxonTerm`s

The ingest renames are breaking for DQM submitters and need advance notice regardless of the rest. Can you confirm these are coordinated?

## On VocabularyTerm vs. the existing `taxon` + `taxon_text` pattern

For **host taxon** I'm satisfied the CV is the right call and I'd merge it as-is. We need the possible entries constrained for later querying and lookup, and `taxon_text` is uncontrolled free text, so it doesn't give us that. I did consider `taxon` + a CV field, but my worry is that anyone asking "which antibodies have mouse as host?" would have to know to check both fields *and* know which values to look for in each. Host species is a closed set of maybe a dozen values, so the `NCBITaxonTerm` FK buys us little here — nobody needs ontology closure over that list, and a dozen display labels are cheap to maintain as term Definitions.

## Proposal: split `antigen_taxon` into two fields

The antigen is a different case, and I don't think it should be collapsed the same way. Antigen source species is an **open** set — potentially any organism whose protein was used as an antigen — which breaks the single-CV approach in two ways: every submitted value has to pre-exist in the CV or the load fails (so either perpetual CV additions on every submission with a new organism, or recurring load errors), and taxonomy-aware queries like "antibodies against any *Drosophila* antigen" start to actually matter. So I'd propose:

- **`antigen_taxon`** — `range: NCBITaxonTerm`; DTO `antigen_taxon_curie`, `range: string`. For the majority case where the NCBITaxon ID is known. This restores the pre-PR names, so there's no breaking change for DQMs on this field.
- **`antigen_taxon_term`** — `range: VocabularyTerm`; DTO `antigen_taxon_term_name`, `range: string`. For everything else.

One constraint is important for this to work: **the two value spaces must be disjoint.** The CV should hold *only* non-taxonomic values (`not specified`, `unknown`, etc.) and never NCBITaxon curies. Then the dual-lookup problem I raised above doesn't arise — a species query is always a single-field query against `antigen_taxon`, and the CV field only records *why* a species is absent. If the CV were allowed to carry curies too, we'd be right back to ambiguity, so that's worth stating explicitly wherever the CV gets defined.

Happy for you to implement, or I can push it. And if curation confirms the antigen list genuinely is closed and small, then the symmetry with `host_taxon_term` is fine and we merge as-is — I'd just like that confirmed rather than assumed.

chris-grove:
We can discuss more details tomorrow

chris-grove:
@kthorat-prog OK, thanks! We actually want to get rid of "antigen_taxon" and "antigen_taxon_curie" entirely after our discussion with Sue this morning. So for antigen taxon we would only ever have "antigen_taxon_term" in the Antibody class and "antigen_taxon_term_name" in the AntibodyDTO class. The slot descriptions probably need to be updated to reflect that. We will stick to the originally suggested CV for antigen (including NCBI Taxon IDs and non)

kthorat-prog:
ohh, my bad. I completely misunderstood. Fixing it now.

chris-grove (approved):
Looks great!

PR #334, opened 2026-09-10
Add BiologicalEntitySet parent and expand Pathway model

sjm41:
## Summary

Introduces an abstract `BiologicalEntitySet` parent to hold the descriptive slots shared by `FunctionalGeneSet` and `Pathway`, and expands `Pathway` from a bare stub (mappings and `id_prefixes` only, no slots, no DTO, no ingest) into a fully submittable class.

Built against the agreed Pathway data class spec. Every field in that spec is accounted for, and the resulting model accommodates **three MODs with no MOD-specific slots**:

- **FlyBase** — FBgg pathway reports, the original driver
- **RGD** — via the Pathway Ontology, which is RGD-maintained
- **SGD** — biochemical pathways map onto `PW:0000002` (classic metabolic pathway); descriptions satisfy the expected summary note

## Pathway-specific additions

| Need | Resolution |
|---|---|
| Pathway type | `pathway_type`, **required**, `range: PWTerm` — one of the five top-level PW terms, verified against OLS |
| Pathway image | `pathway_image_filename`, optional — **interim**, see below |
| GO-CAM model ID | No new slot — submitted as a `cross_reference` with the `go.model` prefix |

`pathway_image_filename` records only a bare filename, with no definition of where files are stored or how a consumer resolves the name. This is deliberate and documented in the slot's `notes:`. The Alliance has no mechanism for submitting or hosting images — `File` in `image.yaml` is a placeholder with no slots, there are no Image/Figure/File DTOs or ingest sets, and none of those classes is implemented in the curation app. To be revisited with the image working group.

## Relation correction

`GenePathwayAssociation` used `subproperty_of: has_participant` (RO:0000057). The association's subject is the gene, so that read *"gene has_participant pathway"* — inverted with respect to the equivalent `is_member_of` on `GeneFunctionalGeneSetAssociation`. Now `participates_in` (RO:0000056).

This predates the PR and was harmless while `Pathway` had no DTO or ingest set. It matters now, because it becomes the relation DQMs are told to use.

## Note for reviewers: this changes FunctionalGeneSet's shipped contract

`FunctionalGeneSet` shipped in v2.17.0 (2026-06-04). Reparenting it changes that contract:

- `symbol` becomes **optional** (`full_name` stays required) — per the spec
- `FunctionalGeneSetSynonymSlotAnnotation` → `BiologicalEntitySetSynonymSlotAnnotation`
- `single_functional_gene_set` → `single_biological_entity_set`
- `parent_sets` is narrowed to the child class via `slot_usage`; `related_sets` stays ranged on `BiologicalEntitySet` so a pathway can relate to a protein complex, as curated at FlyBase

This is safe, but the reason isn't visible in the diff: **`FunctionalGeneSet` was never operationalised at any layer.** No Java entity, no REST endpoint (`/api/functionalgeneset` returns 404 where `/api/gene` returns 401), no loader, no data, and no relation vocabulary. There is nothing to migrate and no re-submission to ask of any MOD.

## Prerequisites before any load

Three items, none visible in the diff, all blocking:

1. Create CV `'Gene Pathway Association Relation'` with term `participates_in`
2. Create CV `'Gene Functional Gene Set Association Relation'` with term `is_member_of` — a pre-existing gap, not introduced here, which blocks `FunctionalGeneSet` loading too
3. Confirm **PW is actually loaded**. `PW` was missing from `ontology_bulk_load_type_enum` and is added here, but `PWTerm` was declared long before without ever being wired to the loader. `pathway_type` is required, so it hard-fails if the terms aren't populated.

## Testing

`test/data/pathway_ingest_test.json` is built from the spec's two worked examples — HEDGEHOG SIGNALING PATHWAY (`FBgg0000978`, signaling) and GLYCOLYSIS (`FBgg0002043`, metabolic) — using real FBgn/FBgg/FBrf identifiers. Between them they exercise `pathway_type`, `pathway_image_filename`, a GO-CAM cross-reference, `parent_sets`, `related_sets` pointing at a protein complex, key GO terms, synonym and note attribution, and 17 gene participation associations. Registered in `SCHEMA_TEST_EXAMPLES`.

Both `gen-json-schema` invocations build clean, and both the pathway and functional gene set ingest tests validate against the closed Ingest schema.

The three `functional_gene_set_*_test.json` files are updated for the slot rename. They are not registered in the Makefile, so were not being validated.

## Not addressed

- **Per-value reference attribution** on `pathway_type`, `set_go_*_terms`, `parent_sets` and `related_sets`. The spec marks these as needing attribution, but always to the same constant internal reference, so association classes to store it would carry no information. Synonyms, notes and gene membership already carry evidence and do so in the test data.
- **Requiring the summary note.** LinkML cannot require a note of a particular type within `related_notes`, so this is enforced at load and documented in `Pathway`'s `notes:`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -207,6 +207,7 @@ SCHEMA_TEST_EXAMPLES := \
 	genome_assembly_test \
 	high_throughput_expression_test \
 	ontology_closure_test \
+	pathway_ingest_test \
 	phenotype_agm_test \
 	phenotype_allele_test \
 	phenotype_gene_test \
--- a/generated/jsonschema/allianceModel.schema.json
+++ b/generated/jsonschema/allianceModel.schema.json
@@ -35666,12 +35666,22 @@
         },
         "ProteinComplex": {
             "additionalProperties": false,
-            "description": "",
+            "description": "A set of proteins that assemble into a stable complex to carry out a biological function.",
             "properties": {
                 "created_by": {
                     "description": "The individual that created the entity.",
                     "type": [
                         "string",
+                        "null"
+                    ]
+                },
+                "cross_references": {
+                    "description": "Holds between an object and its CrossReferences.",
+                    "items": {
+                        "$ref": "#/$defs/CrossReference"
+                    },
+                    "type": [
+                        "array",
                         "null"
                     ]
                 },
@@ -35729,6 +35739,10 @@
                         "null"
                     ]
                 },
+                "full_name": {
+                    "description": "The full name of the entity",
+                    "type": "string"
+                },
                 "internal": {
                     "description": "Classifies the entity as private (for internal use) or not (for public use).",
                     "type": "boolean"
@@ -35744,6 +35758,16 @@
                     "description": "Entity is no longer current.",
                     "type": [
                         "boolean",
+                        "null"
+                    ]
+                },
+                "parent_sets": {
+                    "description": "Parent set(s) of which this set is a subset.",
+                    "items": {
+                        "$ref": "#/$defs/ProteinComplex"
+                    },
+                    "type": [
+                        "array",
                         "null"
                     ]
                 },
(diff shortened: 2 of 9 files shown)


sjm41:
Both assertions were correct and are fixed in d84f6673.

**`related_sets` couldn't hold a `ProteinComplex`** — confirmed, and the diagnosis was exactly right: the widening was pointless because `ProteinComplex` was left on `SubmittedObject`, and the ingest test passed only because `related_set_identifiers` is a bare string. `ProteinComplex` is now `is_a: BiologicalEntitySet`, so `FB:FBgg0001342` on the Hedgehog pathway is genuinely in range. It gains the shared descriptive slots including a required `full_name`, plus `parent_sets` narrowed to `ProteinComplex` for consistency, and a description it previously lacked. Safe for the same reason as the `FunctionalGeneSet` changes — no entity, endpoint, loader or data.

I've left `GeneCluster`, `Operon` and `UniGeneSet` alone. They do share the semantics, but nothing in this PR needs them and they would expand the blast radius further; worth a separate pass.

**`related_note` vs `related_notes`** — confirmed, a straightforward error in the `Pathway` notes. Now reads "exactly one `related_notes` entry of type 'summary'".

---

On the questions — both are for the persistence/loader side and neither is settled. Adding them to the prerequisites already listed in the description:

- **Cross-subclass `primary_external_id` uniqueness.** `BiologicalEntitySet` deliberately carries no uniqueness note, unlike `BiologicalEntity` (`core.yaml:134-137`). FlyBase uses one `FBgg*` namespace for gene groups, pathways *and* complexes, and presents them as distinct concepts, so the same identifier should not appear as two subclasses — but nothing in the schema enforces that today.
- **Resolving `related_set_identifiers`.** Now that three subclasses are resolvable and `related_sets` stays on the abstract parent, the loader has to search across all `BiologicalEntitySet` subclasses. That is the point of the widening, but it makes the uniqueness question above load-bearing rather than theoretical.

Both need an A-Team answer before a load, alongside the two relation vocabularies and confirming PW is loaded.

sjm41:
Link to a comparison table I made (manually):
https://docs.google.com/spreadsheets/d/1DRPWNsy_QWi2lPHKKctyFujXmUvh6aWuW7awIsUJZP0/edit?gid=0#gid=0

PR #335, opened 2026-09-15
Scrum 6452 abc linkML integration

sweng66:


sweng66:
@claude Check this PR?

sweng66:
Thanks to both reviewers — the shape-flip finding in particular was exactly the kind of side effect this change needed caught. Everything actionable is addressed in `4128a82a`; answers to the judgment-call questions follow.

## Fixed in 4128a82a

**Embedded → string flip on 8 entity-side slots (both reviews' headline).** Took the suggested fix: the Agent-level identifier stays, and the slots meant to remain embedded are pinned with explicit `inlined: true` (`+ inlined_as_list: true` for the multivalued `generated_by` / `manufactured_by` / `designating_laboratories`): `data_provider`, `secondary_data_provider`, `affiliated_alliance_member`, `laboratory_of_origin`, `designating_laboratory` as well. I then ran an exhaustive shape diff of **every property of every class** in the generated aggregate against the pre-widening artifact: the only property shape changes remaining are `created_by`/`updated_by` — the agreed goal. (`designating_persons` was checked too; it's `range: string` and was never affected.) This also makes the `secondary_data_provider` description accurate again (assertion 2).

**The rest:**
- `entity_type` (core): explicit `range: string`, no more silent default_range fall-through.
- TET fixtures: `okta-example` replaced with `AGRKB:103000000000001` (valid) / `ACKnowledge_pipeline` (invalids) — fixtures now teach the documented contract.
- ERA fixture `entity_id` values are now curie-prefixed (`SGD:S000002284`, `SGD:S000186503`), matching the description; prefixed is canonical (established in e3071922).
- Third TET rule now has an invalid fixture (`tet_entity_id_validation_without_entity.json`) and a `test-abc` assertion.
- `stage-abc-jsonschema` no longer nests `abc/abc/` on a repeat run; the two abc artifact rules now depend on `$(SOURCE_FILES)` so a local core.yaml edit regenerates them.
- Redundant `required: true` dropped from the `topic_entity_tag_id` / `user_id` identifiers.
- `auth_email`: the optionality is deliberate (identity is the AGRKB curie, not the auth account) — now stated in the slot description instead of implied.
- `Person.unique_keys` description now says plainly that `unique_id` is nullable post-transition; rows that carry it stay unique. We did not add `required: true` — forcing every future Person to carry a legacy concatenated id indefinitely seemed worse than an honest nullable unique key.
- `AutomatedAgent.curie`: `slot_usage` refines the range to `string` — bare stable names are now honestly typed rather than passed through `uriorcurie`. We deliberately chose not to prefix (`AGR:ACKnowledge_pipeline`): the values must match ABC `users.id` verbatim so that no data migration or mapping layer is needed.
- Stale `domain: AuditedObject` dropped from `created_by`/`updated_by` (they're now used by non-AuditedObject ABC classes).

## Answers to the questions

**Shared-slot overloading vs ABC-prefixed slots with `alias:`.** Deliberate, and `alias:` was ruled out on a technical finding: linkml compiles class-rule `if/then` clauses with the slot *name* while properties serialize under the *alias*, so rules on aliased slots silently never fire (verified on linkml 1.9.6 with a minimal schema). Both ABC classes depend on rules. `slot_usage` refinement is also the aggregate's existing convention, and each overloaded reuse states its contextual meaning in the description.

**Generic unprefixed slot names (`topic`, `entity`, `note`, `species`, …).** The ABC column names are the external contract (search-index document keys, MOD submission keys), and per the above, `alias:` can't safely preserve them where rules are involved — so the plain names enter the namespace. None collide today; if a future schema needs `topic` etc. with different semantics, that PR can refine per class the same way.

**Should the ABC classes be in allianceModel at all?** That was the explicit request from the A-team side (Chris G) — resolving the disparate namespaces rather than keeping a standalone `abc/` model. `EntityReferenceAssociationIngest` is deliberately its own root for the ABC pipeline, not registered on `Ingest` — MOD ERA submissions go to the ABC, not through the persistent-store ingest.

**okta → auth rename / Organization curies / unique_id migration path / created_by semantics.** All coordinated with Chris G: the Person→Agent widening and de-Okta were agreed before this PR, with the implementation-side migrations (Java field renames, populating Organization and Person curies, re-keying off `unique_id`) explicitly on the A-team's side ("we can sort out our end once that is in the model"). `created_by_curie` keeps its name — renaming would break every DTO submission; its description now spells out what it accepts and how the loader resolves.

**`AutomatedAgent.evidence_code`.** A DQM-call decision (2026-09-15), not an invention of this PR — the call notes specify an ECOTerm slot on the automated-agent class. It's modeled as a per-agent *default* that an annotation's own evidence code overrides; if it turns out unused, removing it later is cheap.

**ERA `created_by` as plain string, unmapped ids.** Deliberate: unmapped ids are legal and stored verbatim in the ABC (they become automation users) — the `CuratorMapping` class description says exactly this, so `range: CuratorMapping` reference-checking would wrongly reject legal submissions.

**`data_novelty` / `secondary_data_provider_abbreviation` NOT NULL.** Verified directly against the refreshed literature-4006 database during this work: `data_novelty` and `data_context` are `NOT NULL` on `topic_entity_tag`; `secondary_data_provider_id` is `NOT NULL` on `topic_entity_tag_source`. Search-index documents do carry the internal ABC primary keys (`topic_entity_tag_id`, `topic_entity_tag_source_id`); `reference_id` is the one relaxed to optional because public consumers key on `reference_curie`.

**Committed artifacts lagging head.** Correct — CI's regeneration runs per push and has now caught up; the artifact diff for `4128a82a` should show `created_by`/`updated_by` as the only shape change across the aggregate.

